PR #6, opened 2021-01-17
Fix warnings in unit tests

Arun-Niranjan:
This PR adds the optional `dtype=object` argument to `np.array(...)` when the input lists/arrays are of different dimensions
in order to suppress warnings when running tests.

We should revisit how arrays of different dimensions are handled in general, it is likely that an abstraction exists which will be safer and easier to use/test

### DIFF
--- a/test/test_categorical.py
+++ b/test/test_categorical.py
@@ -50,7 +50,7 @@ class TestCategorical(unittest.TestCase):
     def test_multi_factor_init_values(self):
         values_1 = np.random.rand(5, 4)
         values_2 = np.random.rand(4, 3)
-        values = np.array([values_1, values_2])
+        values = np.array([values_1, values_2], dtype=object)
         c = Categorical(values=values)
         self.assertEqual(c.shape, (2,))
         self.assertEqual(c[0].shape, (5, 4))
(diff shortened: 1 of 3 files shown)


conorheins:
We've thought a lot about how to deal with 'ragged' arrays in this context, and numpy object arrays (or as we've been colloquially calling them, AoAs or 'arrays-of-arrays') seem to be the best option that we know of (better than dicts or lists, for instance). Would be happy to hear any further suggestions though, I'm by no means 100% convinced that AoAs has to be the way to go.

conorheins (approved):
Just testing out the 'review changes' functionality to get the hang of it & approving here.

Arun-Niranjan:
I think the issue is whether anything that uses an array that _may or may not have_ sub arrays of varying length has to check the dimensions of everything are compatible for whatever it needs to do. I imagine this could/has lead to a lot of boiler plate code checking this.

I would suggest a thin wrapper class for AoAs that has validation methods on it - then the user can just call a one-liner method and know immediately that everything is happy before proceeding.

However I'd say this is a low priority for now.

conorheins:
> I think the issue is whether anything that uses an array that _may or may not have_ sub arrays of varying length has to check the dimensions of everything are compatible for whatever it needs to do. I imagine this could/has lead to a lot of boiler plate code checking this.

That's a good point - and you're right, this has caused issues & a ton of boilerplate calling of utility functions that basically do this kind of validation / dimension-checking / dimension extraction.

> 
> I would suggest a thin wrapper class for AoAs that has validation methods on it - then the user can just call a one-liner method and know immediately that everything is happy before proceeding.

I like that idea! I agree about it being low-priority, but let's definitely revisit this :)

alec-tschantz:
> I would suggest a thin wrapper class for AoAs that has validation methods on it - then the user can just call a one-liner method and know immediately that everything is happy before proceeding.

I agree - we do have an "obj_array" function in utils for this exact purpose, but it doesn't do much at the moment (just initialises an object array of some given shape). Would be happy to put all validation here

alec-tschantz (approved):

PR #7, opened 2021-01-24
Fix import in test_control.py

Arun-Niranjan:
This import fix gets the entire test suite green. Will try and untangle other dependencies in a separate PR.


Arun-Niranjan:
tagging @conorheins @alec-tschantz for review

conorheins:
Great, looks like it was another vestige from when we didn't have all the sub-modules of core factored out, as they are now.

PR #8, opened 2021-01-24
Add ci action

Arun-Niranjan:
Now whenever someone creates a PR,  pymdp unit tests will run automatically using python 3.8 and the modules listed in requirements.txt in a clean test environment

Note this currently ignores the experimental folder.

I am treating any flake8 linting problems as warnings only for now - but ideally we find a linter we're happy with and eventually fail builds which don't pass the linter.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.9]
+        python-version: [3.8]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -1,0 +1,37 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Python package
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.9]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install flake8 pytest
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+    - name: Lint with flake8
+      run: |
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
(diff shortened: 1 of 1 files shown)


Arun-Niranjan:
tagging @conorheins and @alec-tschantz for review :-)

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,37 @@
...
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.8]
```
Arun-Niranjan, line 18: interestingly, our unit tests fail on python 3.9 (I think you can see the failed [build logs](https://github.com/Arun-Niranjan/pymdp/pull/3/checks?sha=9b38878948bb6ed7be926a5a4ec18747c06ed775) for the previous commit on this branch)

> E     File "/home/runner/work/pymdp/pymdp/test/test_control.py", line 28
> p = np.random.rand((*([no] + n_states)))
> E                             ^
> E   SyntaxError: can't use starred expression here

I think this should be a straightforward fix, but it should be done in a separate PR.

conorheins: Thanks, nice find. I need to look at Python 3.9 to see what the exact restrictions are, but  the 'official' numpy.random documentation recommends using `np.random.random_sample()` when passing in variable shape tuples (see https://numpy.org/devdocs/reference/random/generated/numpy.random.rand.html), so I will change those across the board when we're initializing random uniform matrices.

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,37 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Python package
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
Arun-Niranjan, line 10: this might need some tweaking, but the easiest way to experiment is to land this PR as it is and push/raise Rs

conorheins: Cool. So I merged your commit, pulled and then pushed a new change (just replacing those calls to `np.random.rand` that you found in `test_control`), but I didn't see anything reflecting that the tests were run.

Arun-Niranjan: hmm - if you are directly pushing to master I would expect those to run.

Although best practise is to not push directly to master, try fixing something else and creating a PR? We'll see if that runs.

conorheins: Yeah I've just been pushing because I've (embarrassingly) never really worked with separate feature branches / issuing pull requests - I've always just been pushing directly to master. I'll do that now though to test this out - it's high time I learned how to use PRs properly anyway :)

Arun-Niranjan: It is honestly wonderful once you get the hang of it :-) I highly recommend reading this: https://github.com/susam/gitpr#create-pull-request - it's what I've been using for contributing to pymdp so far!

conorheins: This looks like exactly what I need, thank you! So I just created a branch, made a tiny change to the control.py file and then made a pull request. It did all the tests, as intended :) Super cool

PR #10, opened 2021-01-24
Tidy up demo .py files

Arun-Niranjan:
General tidy up to get the .py files in experimental/examples to run.

I have just about managed to get jupyter notebooks to run and visualise in the windows linux susbsytem, we can address those in a separate PR.


Arun-Niranjan:
A quick note - the current CI pipeline excludes the experimental folder - but I ran the affected files locally and the plots seemed to work (even if I don't fully understand the maths yet)

PR #13, opened 2021-01-31
Added computation of policy VFE in case of fixed point MMP 

conorheins:
I don't have a way of validating this computation yet (any ideas @alec-tschantz @BerenMillidge), since spm_MDP_VB_X doesn't use fixed points when updating beliefs, but the computation of the free energy "works" in terms of argument-handling. But would be good to get a second pair of eyes on that / discuss how we can make sure the maths is correct.

### DIFF
--- a/pymdp/core/algos/mmp_v2.py
+++ b/pymdp/core/algos/mmp_v2.py
@@ -11,7 +11,7 @@ __author__: Conor Heins, Beren Millidge, Alexander Tschantz, Brennan Klein
 import numpy as np
 
 from pymdp.core.utils import to_arr_of_arr, get_model_dimensions, obj_array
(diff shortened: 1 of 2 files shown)


MagnusKoudahl:
Is this for Meanfield marginal message passing? In that case a useful diagnostic we use in my current lab is to examine the FE after each MP iteration. We can prove that FE can never increase after message passing. So if it ever does anything but decrease it's either a numerical error or a bug.

conorheins:
> Is this for Meanfield marginal message passing? In that case a useful diagnostic we use in my current lab is to examine the FE after each MP iteration. We can prove that FE can never increase after message passing. So if it ever does anything but decrease it's either a numerical error or a bug.

This is great @MagnusKoudahl thanks for the tip! I'll definitely add that in as a check then and credit you :)

PR #56, opened 2021-12-15
Proposed fixes to control.py and utils.py

spetey:
1) control.py's action_marginals should sum only over the first time step, 2) utils.py's sample(probabilities) throws an error with the argument "squeezed"


conorheins:
thanks for this @spetey -- before I run the CI test-suite, can you please remove the extra files from your commit (those ones committed besides `utils.py` and `control.py`?

spetey:
Ugh, I've tried to reset to HEAD~1 and ignore the "egg-info" directory that
was auto-added somehow, but then there were conflicts because the upstream
repo had changed, and then it stopped ignoring the "egg-info" ... I find
git terribly confusing, I tried to fix it as best I could, and opened
another pull request. Thanks again for your patience.

Steve

conorheins:
Hey @spetey I've pushed another branch for the documentation that also included the fix to `control.py` already, so we don't need that commit anymore. So all that remains is to figure out the `utils.sample()` function, that you're proposing a change to. Let me know what the error is and we can figure out whether dropping that `.squeeze()` is the right way forward.

alec-tschantz:
> _Ugh, I've tried to reset to HEAD~1 and ignore the "egg-info" directory that was auto-added somehow, but then there were conflicts because the upstream repo had changed, and then it stopped ignoring the "egg-info" ... I find git terribly confusing, I tried to fix it as best I could, and opened another pull request._ 

I've added a gitignore *.egg.info to a PR, so this won't be an issue going forward

PR #65, opened 2022-01-17
numpy subclass proposal

alec-tschantz:
This is a conversation piece but spent some time looking into subclassing NumPy arrays and I think we can get everything we need, including the nuanced object structure. Here is a basic example

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,4 @@ __pycache__
 .ipynb_checkpoints/
 .pytest_cache
 env/
-pymdp.egg-info
+*.egg-info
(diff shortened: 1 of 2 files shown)


conorheins:
I like the idea of replacing the object arrays -- they're not that bad to work with but I agree having our own type (something like the original `Categorical` class you made way-back-when, but more flexible/not as linked to distributions per se, would be idea.

I see this as something that we could put into the v1.0.0 release, since it'd require some deeper refactoring and changing of documentation, etc.

Btw I've left some comments in the example because I'm a little unsure how you're doing object arrays here.

conorheins:
@alec-tschantz see my suggested changes in d6a56d6

PR #95, opened 2022-08-27
fix plot_likelihood in some examples

tverbele:
`plot_likelihood` is missing a row in some of the examples?

### DIFF
--- a/docs/notebooks/tmaze_demo.ipynb
+++ b/docs/notebooks/tmaze_demo.ipynb
@@ -81,8 +81,8 @@
     "    \n",
     "def plot_likelihood(A, title=\"\"):\n",
     "    ax = sns.heatmap(A, cmap=\"OrRd\", linewidth=2.5)\n",
-    "    plt.xticks(range(A.shape[1]))\n",
-    "    plt.yticks(range(A.shape[0]))\n",
+    "    plt.xticks(range(A.shape[1] + 1))\n",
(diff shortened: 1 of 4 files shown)


conorheins:
Hi @tverbele -- maybe it's something to do with my version of seaborn/Python, but I don't really see any difference in whether you increase the number of xticks/yticks by one? See the two images below for examples of the plot outputs with/without your changes.

Without:
image

With:
image

Basically the only difference I can see between Without and With, is that there is an extra tick at the end of each axis (bottom left and bottom right, respectively) with your change. Is there a reason /some context where think this is helpful / necessary?

tverbele:
When I execute the `tmaze_demo` example I get with master:



whereas with my patch



Which version of matplotlib / seaborn are you using?
I'm testing with
matplotlib==3.2.2
seaborn==0.11.2

Regardless, it might be a good to move the plot_xxx functions to `utils` or another pymdp module instead of re-defining them in multiple examples?

tverbele:
Ok, I also checked on a different machine with a more recent matplotlib version (3.5.2), and there it does work correctly ... so probably just a bug in matplotlib.

conorheins:
Hey @tverbele, as we discussed, although you closed this PR/deleted the branch, I'm going to start an issue / reference this PR and will move the plot_xxx functions to `utils` or another module to address this issue, while also adding in the explicit axes arguments to facilitate compatibility with earlier matplotlib (e.g. 3.2.2)

Issue #115, opened 2023-05-24
Choose random action if all action probabilities are the same, even if sampling is deterministic (i.e., some tolerance of probability)


conorheins:
When `action_selection == "determinstic"` in `control.sample_action()`, but all the action probabilities are equal, we should sample rather than deterministically choose the first action (which is the default behavior of `np.argmax`

conorheins:
Adressed this [here](https://github.com/infer-actively/pymdp/pull/116), now closing

conorheins:
This has been better addressed and dealt with by @AleMuzzi in [this pull request](https://github.com/infer-actively/pymdp/pull/117), dealing specifically with the case when there is a subset of the total number of actions that have equal probability, deterministic sampling still needs to sample from among them

PR #116, opened 2023-05-24
Deterministic action selection 115

conorheins:
Address special case of equi-probable actions even with `action_selection = "deterministic"` in `control`, to address Issue #115


conorheins:
Thanks to @jkbren @daphnedemekas @AleMuzzi for pointing this out

PR #129, opened 2024-01-20
Add complex action dependencies to Sparse likelihoods 111

ran-wei-verses:
Introduced a new `B_factor_control_list` variable so that the user can specify the control variables each transition factor depends on using a list similar to the `B_factor_list` introduced to specify state dependencies.

* Made edits to `utils, control, learning, agent`
* The shape of B matrices is now defined as: `[state_dim] + lagging_state_dims + control_dims`. Conditioning on actions can be done easily with `B[..., *relevant_action_idx]`
* Added checks/assertions in `agent.__init__`
* Added relevant tests. Fixed a bunch of failed tests due to using `!=` rather than `is not` in `agent.__init__`. 
* All tests passed except for those in `test_mmp.py` and `test_SPM_validation.py`

### DIFF
--- a/pymdp/agent.py
+++ b/pymdp/agent.py
@@ -64,8 +64,12 @@ class Agent(object):
         policy_sep_prior=False,
         save_belief_hist=False,
         A_factor_list=None,
-        B_factor_list=None
+        B_factor_list=None,
+        B_factor_control_list=None,
     ):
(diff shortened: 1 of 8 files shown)


dimarkov:
I am closing this as it should have resolved in a different branch.

Issue #130, opened 2024-05-07
Fix `pymdp.utils.obj_array_from_list` 


conorheins:
Creating an object array from a list of numpy arrays seems to break if the component arrays of the list have matching leading axis. For example

```
arrs = [np.zeros((3, 6)), np.zeros((3, 4, 5))]
obj_arr = np.array(arrs, dtype=object)
```

will throw the following error:

```
ValueError: could not broadcast input array from shape (3,6) into shape (3,)
```

whereas 

```
arrs = [np.zeros((4, 6)), np.zeros((3, 4, 5))]
obj_arr = np.array(arrs, dtype=object)
```

works just fine.

PR #131, opened 2024-05-08
make safe obj_array_from_list to address issue #130 with test

ran-wei-verses:
Fix issue #130 where numpy throws broadcasting error while trying to turn a list of arrays with same leading dimensions into an object array. 

Added test for this case.


conorheins:
This looks perfect, thanks for being thorough with the unit test as well :)

PR #137, opened 2024-06-12
Feature/agent demo

toonvdm:
The generalized T-maze, where there now can be multiple cues and distractor in `envs/generalized_tmaze.py`. An example notebook is in `examples/generalized_tmaze_demo.py`


dimarkov (commented):
Looks good! Nice that you added the rollout function.

PR #140, opened 2024-06-12
Fix tests add coverage

scott-carroll-verses-ai:


tverbele:
@scott-carroll-verses-ai why should we hack the `sys.path` for local test execution as opposed to installing the library locally using `pip install -e .` and then running pytest?

scott-carroll-verses-ai:
> @scott-carroll-verses-ai why should we hack the `sys.path` for local test execution as opposed to installing the library locally using `pip install -e .` and then running pytest?

Good question @tverbele . The idea is that you test the code before you install it. Can help with issues during dev re: e.g. needing to wipe out cache and change versions of the lib as you work on the "next" version. i.e. to avoid installing a hacky implementation before testing. YMMV 🤷‍♀️ 

Happy to chat more but my feelings won't be hurt if you decide to not test that way.

Arun-Niranjan:
@scott-carroll-verses-ai I think given how old this PR is and how much merge conflicts there are, I might take parts of this and put them in a new PR and close this one out if you don't mind?

Arun-Niranjan:
closing in favour of https://github.com/infer-actively/pymdp/pull/332

PR #144, opened 2024-06-13
Fix/error c tensor

toonvdm:
This PR resolve the issue that C, D, or E are overwritten if `apply_batch` is True, but C, D, or E are not none.


conorheins (approved):
This looks good to me

PR #146, opened 2024-06-17
Ovf with example

dimarkov:
So this is a new PR containing only changes relevant for the OVF implementation. I have removed the dependence on multimethod in inference.py as I was getting jitting issues.


conorheins (approved):
This looks very nice - thanks for adding these features

PR #147, opened 2024-06-17
Fixing batching over sparse arrays

dimarkov:
I have made possible batching over sparse arrays with vmap, and also merged into single function both sparse and dense array computations for backward smoothing.


conorheins (approved):
Looks good! @toonvdm see here -- we should adapt this version for your unit tests

PR #149, opened 2024-08-05
Add some extra features to the Distribution API 

tverbele:
- add a `Model` class which is the result of `compile_model`
- the model A,B can now be indexed by the observation/state names instead of indices
- the model now also has C and D that can be set via string indexing
- create an agent using `Agent(**model)` which unwraps the model to `A=As, B=Bs, C=Cs, D=Ds`


alec-tschantz (approved):
looks good !

conorheins (approved):

PR #163, opened 2025-01-15
Tmaze a b learning

riddhipits:
Creating a tmaze tutorial script for using jax pymdp and fixed A and B learning such that they can occur together and separately


.gitignore
```diff
@@ -9,3 +9,4 @@ __pycache__
 env/
 pymdp.egg-info
 inferactively_pymdp.egg-info
+venv/
```
conorheins: Did you propose to add this into the overall pymdp repo's gitignore, or is this just a legacy commit from your personal workspace/setup when developing the tmaze notebooks?

pymdp/envs/env.py
```diff
@@ -5,21 +5,28 @@
 from equinox import Module, field, tree_at
 from jax import vmap, random as jr, tree_util as jtu
 import jax.numpy as jnp
+import matplotlib.pyplot as plt
```
conorheins: Could you remove these imports if they're not being used anywhere? (namely, matplotlib.pyplot and numpy)

pymdp/envs/rollout.py
```diff
@@ -2,6 +2,7 @@
 import jax.random as jr
 import jax.tree_util as jtu
 import jax.lax
+import equinox as eqx
```
conorheins: Can you remove this import if it's not beign used anywhere?

.gitignore
riddhipits: Proposing to add it to the overall pymdp repo's .gitignore, as venv is included in default Python .gitignore templates. Including it may also help avoid potential issues with accidentally committing local virtual environments in the future.

pymdp/envs/env.py
riddhipits: I've removed the unused imports as requested. Thanks for pointing that out!

pymdp/envs/rollout.py
riddhipits: I've removed this unused import - thanks for pointing that out!

PR #168, opened 2025-01-27
Fix inference methods comparison notebook

tverbele:
Fix the notebook with v1.0.0 and record the expected outcome of the cells.


conorheins (approved):
Looks good, thank you for fixing :)

PR #169, opened 2025-02-05
remove jax from import

OzanCatalVerses:
`si.py` was still refering to the `pymdp.jax.control` version of `control.py` breaking imports of sophisticated inference. Also removes unused imports.

### DIFF
--- a/pymdp/planning/si.py
+++ b/pymdp/planning/si.py
@@ -1,17 +1,15 @@
 import itertools
-import jax
 import jax.numpy as jnp
 import jax.tree_util as jtu
 from jax import nn
 from jax import vmap
 
 import pymdp
-from pymdp.jax.control import (
+from pymdp.control import (
     compute_info_gain,
     compute_expected_utility,
     compute_expected_state,
     compute_expected_obs,
-    calc_inductive_value_t,
 )
 
 
@@ -56,7 +54,11 @@ class Tree:
         return node["parent"]
 
     def leaves(self):
-        return [node for node in self.nodes if "qs" in node.keys() and len(node["children"]) == 0]
+        return [
+            node
+            for node in self.nodes
+            if "qs" in node.keys() and len(node["children"]) == 0
+        ]
 
     def append(self, node):
         self.nodes.append(node)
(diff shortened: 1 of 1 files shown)


pymdp/planning/si.py
```diff
@@ -269,7 +269,7 @@ def tree_backward(node, prune_penalty=512, gamma=1):
 
 
 def policy_entropy(node):
-    return -jnp.dot(node["q_pi"], jnp.log(node["q_pi"] + pymdp.maths.EPS_VAL))
+    return -jnp.dot(node["q_pi"], jnp.log(node["q_pi"] + 1e-16))
```
conorheins: This may be fixable by simply doing

```
-jnp.dot(node["q_pi"], pymdp.maths.stablelog(node["q_pi"]))`
```

or even

```
-pymdp.maths.stable_xlogx(node["q_pi"]).sum()
```

conorheins:
> Yup, good catch ... @conorheins do we have a new equivalent of the old pymdp.maths.EPS_VAL instead of hard coding 1e-16? We might also want to update the notebooks though...

See my comment above on the policy entropy function -- we avoid evaluating log of 0.0 within the the stable_log function now, because that function clips the output to a tiny value determined by jnp.finfo(float).eps

conorheins (changes requested):
Can you replace that call to jnp.log(node["q_pi"] + 1e-16) with the `stablelog` call?

pymdp/planning/si.py
conorheins: @tverbele was there a reason not to use the stable_xlogx variant here?

pymdp/planning/mcts.py
```diff
@@ -1,12 +1,19 @@
 from functools import partial
 from jax import vmap, nn, random as jr, tree_util as jtu, lax
-from pymdp.jax.control import compute_expected_state, compute_expected_obs, compute_info_gain, compute_expected_utility
+from pymdp.control import (
```
conorheins, line 3: To avoid these formatting diffs I suggest before merging to v1.0.0 alpha / master, we add in some kind of CI workflow step with `black` and a pre-commit hook -- see https://github.com/infer-actively/pymdp/issues/183

pymdp/planning/si.py
tverbele: replaced by `stable_entropy()` now

PR #181, opened 2025-06-04
Chore/cleanup tmaze demo

conorheins:
- add `nb` optional dependencies in `pyproject.toml` which adds visualization-enabling dependencies
- clean up of the `tmaze_demo.ipynb`
- add an `event_dim` argument to `pymdp.maths.expected_value_dirichlet()` which determines which axis to normalize the dirichlet tensor along when computing the expected value


conorheins:
Besides any edits / reviews you may have of the notebook itself @riddhipits , this is the only outstanding issue

Do we want to include those 4 other experiments you listed at the bottom of the initial markdown cell where you're explaining the contents of the notebook

#### **4.** A deterministic generative process, and a single agent solving the task with sophisticated inference

#### **5.** A deterministic generative process, and a single agent solving the task with inductive inference

#### **6.** A deterministic generative process, and multiple agents solving the task with standard active inference

#### **7.** A deterministic generative process, and multiple agents solving the task with sophisticated inference

Including those may be overkill for this notebook, so one easy option would just be to include those special cases (sophisticated + inductive inference, with one and multiple agents) in a separate set of example notebooks, and just remove those markdown bits from here. What do you think?

PR #187, opened 2025-07-01
setup a ruff lint workflow

OzanCatalVerses:
Adds a github action workflow to lint on PR and adds the required config to the pyproject.toml.

@conorheins @tverbele  I do not know how to test this. As it stands this will generate 120-ish warnings, which we can fix by running ruff check fix and committing that. I don't know if that is something we want to do now though.


OzanCatalVerses:
Ok it seems all the checks pass because there are no changes to any python files.

Issue #191, opened 2025-07-25
Implementing exact computation of expected information gain about parameters, i.e. novelty, in pymdp


lancelotdacosta:
The current derivation of novelty which is used in `pymdp` and SPM relies on an asymptotic approximation of the digamma function. This asymptotic approximation is not always verified in practice, especially for posterior Dirichlet counts which are low, i.e. at the beginning of learning. See mathematical details in [infogain_pymdp.pdf](https://github.com/user-attachments/files/21427546/infogain_pymdp.pdf), the reference for the derivations is Appendix D.3 Da Costa et al. ‘Active inference on discrete state-spaces: A synthesis’, Journal of Mathematical Psychology, 2020, and I believe the same derivation is in the Active Inference book by Parr, Pezzulo, and Friston. This suggests that the agent's exploratory behavior at the beginning of learning can be suboptimal or skewed in unexpected ways.

conorheins:
Thanks for this clear explanation @lancelotdacosta and the document! I have assigned the issue to you

conorheins:
Addressed by #193 -- upon review and merging will close

conorheins:
Now addressed by #193

PR #193, opened 2025-07-25
Fixed param info gain computation and created tests - Fixes Issue #191

lancelotdacosta:
Addresses #191 

### Key improvements:
- Replaced the legacy heuristic with exact parameter information gain computation while maintaining backward compatibility and providing comprehensive test coverage.

### 1. `pymdp/maths.py` changes:

- **Added imports**: `digamma` from `jax.scipy.special` 
- **New function `_exact_wnorm`**: Implements exact parameter info gain computation according to eq. (D.15) from Da Costa et al. (2020) with:
  - Single clipping strategy for numerical stability 
  - Direct use of `jnp.log()` and `digamma()` on clipped values
  - **Historical minus sign for backward compatibility**
- **Modified `spm_wnorm`**: 
  - Updated docstring to explain the exact vs legacy toggle
  - Added conditional logic to use `_exact_wnorm()` when `exact_param_info_gain` argument is set to `True`
  - Preserved legacy heuristic as fallback

### 2. New file `test/test_param_info_gain_jax.py` (115 lines):

- **`test_exact_wnorm_finite`**: Numerical stability test across extreme scales (1e-12 to 1e12)
- **`test_exact_wnorm_mathematical_correctness`**: Algebraic verification using hand-computed 2×2 example  
- **`test_calc_pA_info_gain_precision`**: Behavioral test verifying monotonic relationship between precision of beliefs and negative information gain using T-Maze dimensions, where precisions are randomized across 5 seeds

### DIFF
--- a/pymdp/maths.py
+++ b/pymdp/maths.py
@@ -3,7 +3,7 @@ import jax.numpy as jnp
 from functools import partial
 from typing import Optional, Tuple, List
 from jax import tree_util, nn, jit, vmap, lax
-from jax.scipy.special import xlogy
+from jax.scipy.special import xlogy, digamma
 from opt_einsum import contract
 from multimethod import multimethod
 from jaxtyping import ArrayLike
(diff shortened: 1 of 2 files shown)


pymdp/maths.py
```diff
@@ -12,6 +12,13 @@
 
 MINVAL = jnp.finfo(float).eps
 
+# --- Toggle for information-gain formulation ---------------------------------
+# Set to ``True`` to use exact information-gain over parameters instead of the legacy
+# spm_wnorm heuristic.  Changing this single line lets you switch behaviour
+# globally without touching the rest of the code-base.
+USE_EXACT_PARAM_INFO_GAIN = True  # ⇦ CHANGE THIS LINE TO ``True`` FOR THE FIX
```
conorheins: Instead of having this toggle be a global variable defined at the top of the function, I suggest it is an argument to `spm_wnorm(..., use_exact=True)` that defaults to True. This way we can easily functionally change between the exact vs. approximate version without having to change a global variable defined within a file

lancelotdacosta:
Okay, I just fixed this as requested!

conorheins (approved):
Looks great! This is an exemplary PR :)

I updated PR description with the latest change you made to remove reference to that global variable

PR #214, opened 2025-08-05
Fix/rollout

conorheins:
Addreses #215 
This PR (by @tverbele) refactors the `rollout()` function to call a separate `infer_and_plan()` function that runs a single timestep of hidden state inference and planning.

It also updates the following graph worlds notebooks to use this new form of the rollout function:
- `examples/envs/graph_worlds_demo.ipynb`
- `examples/experimental/sophisticated_inference/si_graph_world.ipynb`

After this PR is merged, other notebooks will have to be updated in the `chore/notebook_cleanup` branch.

It also fixes the sophisticated inference implementation in `si.py` to work with `infer_and_plan()`.

Drive-by commits:
- sets `learn_A`, `learn_B`, and `learn_D` to `False` by default in the `Agent` constructor


conorheins:
I will review since I'm just opening this PR on @tverbele 's behalf

Arun-Niranjan (approved):

pymdp/envs/rollout.py
```diff
@@ -34,134 +104,75 @@ def rollout(agent: Agent, env: Env, num_timesteps: int, rng_key: jr.PRNGKey, pol
...
-            empirical_prior=empirical_prior,
-        )
+        keys = jr.split(rng_key, batch_size + 2)
+        rng_key = keys[0] # carry first key
 
-        rng_key, key = jr.split(rng_key)
-        qpi, _ = policy_search(agent, qs, key) # compute policy posterior using EFE - uses C to consider preferred outcomes
+        # infer next action and beliefs using the agent's inference and planning methods
+        # first timestep we don't have a prev action yet, so then we call with action=None
+        action_next, qs, xtra = jax.lax.cond(t > 0, lambda : infer_and_plan(agent, qs_prev, observation, action, keys[1], policy_search=policy_search), lambda : infer_and_plan(agent, qs_prev, observation, None, keys[1], policy_search=policy_search))
```
tverbele: perhaps we shouldn't check on `t` value, as one might want to provide a valid action with init_carry, and use this to calculate an empirical prior instead of agent.D ? :thinking:

Arun-Niranjan: would that be fixed by setting `t>=0`?

given the overall change was necessary to fix up the notebooks, I'd suggest we land and revisit this question in a separate PR (ideally with unit tests demonstrating the desired behaviour)

tverbele: addressed in my next commit

PR #217, opened 2025-08-11
allow to call infer_states with onehot_obs=True

tverbele:
this is handy if you want to update with a distribution instead of an integer, without setting the whole Agent object to onehot_obs=True


conorheins (approved):
Looks good to me

PR #218, opened 2025-08-11
adding epsilon to tensor distribution and then normalising to avoid division by zero

riddhipits:


Arun-Niranjan:
For my understanding @riddhipits, would this occur if a user has incorrectly specified their initial generative model? Or could it result from a future evolution during training/learning?

Arun-Niranjan:
That sounds good to me. I had two principles in mind:
1. Fail fast (as you described)
2. Third party libraries should fail gracefully where possible, especially due to internal processes.

In this case I think 1. outranks 2., and I like how you've structured the guard clause.

conorheins (changes requested):
Here's my take:

**Warnings about un-normalized categorical distirbutions**
I totally support the idea of throwing warnings / errors about these 'sum-to-zero' columns. I think we should even throw a more generic warning about anything that doesn't sum to 1.0 (which will include both these 0.0-filled columns of probability tensors, as well as other cases which are non-zero but not normalized). However, I think this should be done outside the `norm_dist` function, which should remain a (potentially jittable) primitive operation that doesn't raise any errors inside -- see the problems with doing so [here](https://github.com/jax-ml/jax/issues/4257). I think it could rather be done either
- in the `self._validate()` method of the `Agent` class, see [here](https://github.com/infer-actively/pymdp/blob/chore/notebook_cleanup/pymdp/agent.py#L699).
- in a separate function that the `self._validate()` method could itself call, but which could also be used separately in other parts of the `Agent` class's `__init__`  or even standalone on any conditional probability tensor whose columns we expect to sum to 1.0

**Filling zero-valued columns with small values**
I think performing this filling shouldn't be the default behavior of any function, because it's actually changing the mathematical parameterization of the generative model. In doing `adjusted_dist = jnp.where(zero_sum_mask, epsilon_value, original_dist)` you're assuming that a particular conditional distribution p(X | Y, Z, ...) of your model is a uniform categorical distribution with probabilities equal to $\frac{1}{K}$, where $K = |X|$. Doing this may have unintended side-effects like leading to concentration of probability mass in expected places during message passing, arbitrary inflations of information gain terms (due to the manipulation of particular marginal entropies e.g. in the state information gain) and also leaves the specification of the corresponding Dirichlet prior over that distribution ambiguous (since there are infinite Dirichlet pseudocount scales that may lead to an expected value of the flat categorical distribution).

Given this, I propose we leave any alterations of the distributions themselves to be on the onus of the user, but we provide warnings/errors and can even output the indices (the equivalent of the `zero_sum_mask`) of the problematic distributions with these validation functions. But then the user should be the one to adjust their distribution and makes choices about the Dirichlet priors, etc.

pymdp/distribution.py
```diff
@@ -90,8 +90,8 @@ def __setitem__(self, indices, value):
         ]
         self.data[tuple(index_list)] = value
 
-    def normalize(self):
-        self.data = norm_dist(self.data)
+    def normalize(self, add_noise: float = 0.0):
```
conorheins: Aside from my argument that `add_noise` shouldn't be part of this function, I also think it should be called `epsilon` or something since the term `noise` suggests there's some sampling happening under the hood.

Arun-Niranjan:
thanks for the detailed response @conorheins - moving this into the Agent `_validate()` method sounds like a good way forward.

I hadn't considered the error handling within JAX - that is very useful context indeed. It sounds like we have to keep our validation and execution code entirely separate at the function level in order to make the most of JAX (I dislike the callback solution mentioned in that issue, that's sounds like unnecessary complexity)

long term I suggest we make the `Agent` class composable, such that on construction the agent can be passing in a function or object that performs the validation required (with a sensible default). "Carve it at the joints" to steal a phrase.

This would: 
- explicitly keep any code that we expect to throw errors off the JITtable hot-path
- make that code more easily testable and signal to users (and contributors!) that's this is where exception-raising code should go
- allow users to override the validation easily if they want to experiment

Arun-Niranjan:
This is looking a lot better - @riddhipits if you could add a few unit tests in https://github.com/infer-actively/pymdp/blob/chore/notebook_cleanup/test/test_utils.py that would be great! 🎉

Issue #225, opened 2025-08-20
Ensure that `examples/envs/generalized_tmaze_demo.ipynb` works with `rollout` function and `batch_size` arg


conorheins:
This notebook needs to be run again to ensure that it works with the latest version of `rollout` (after the refactor introduced by #214) and updated based on the new `batch_size` argument to `Agent`

PR #226, opened 2025-08-21
Fixed the rollout in the `mcts_generalized_tmaze` example

salvatomm:
Have fixed the notebook with the new rollout. Now it runs, gives the expected results, and renders them well.


conorheins:
This looks good! Can you also fix the one mentioned in #225, namely in `examples/envs/generalized_tmaze_demo.ipynb`? I am not sure how similar they are but I assume they use the same environment class

PR #230, opened 2025-08-22
update `examples/api/distribution_api.ipynb` with `batch_size` argument to `Agent`

salvatomm:
Addressed the distribution_api file. It now runs properly with the batching


conorheins:
Awesome, thanks a lot @salvatomm !

Issue #247, opened 2025-09-03
FrozenInstanceError when setting environmental state in `examples/envs/generalized_tmaze_demo.ipynb`


conorheins:
When going through the notebooks to ensure backcompatbility after merge of #234, I ran into the following FrozenInstanceError when rendering the sequence of environmental states after calling rollout in the generalized TMaze demo notebook:

```python
images = []
for t in range(T):
    env_state = copy(info['env'])
    env_state.state = jtu.tree_map(lambda x: x[:, t], info['env'].state)
    images.append(np.array(render(env_info_m, env_state, show_img=False)) )

---------------------------------------------------------------------------
FrozenInstanceError                       Traceback (most recent call last)
Cell In[6], line 4
      2 for t in range(T):
      3     env_state = copy(info['env'])
----> 4     env_state.state = jtu.tree_map(lambda x: x[:, t], info['env'].state)
      5     images.append(np.array(render(env_info_m, env_state, show_img=False)) )

File <string>:4, in __setattr__(self, name, value)

FrozenInstanceError: cannot assign to field 'state'
```

I believe, but have not verified, that @salvatomm was not hitting this error when he updated the notebook in his PR #228 due to a version mismatch in his local dev versions of jax and/or equinox.

This should be fixed with the use of `eqx.tree_at()` instead of an explicit assignment. 

A more general note: I assume setter methods like `__setattr__` method on FrozenDicts and Equinox modules is  less stable across jax and equinox versions and we can avoid this by using `eqx.tree_at(...)`` and `at[idx].set(val)` notation when in doubt.

PR #248, opened 2025-09-03
Fix `generalized_tmaze_demo.ipynb` with more stable tree_at notation

conorheins:
Addresses #247 

- use `eqx.tree_at()` to update environmental state with time-indexed state before image rendering, to avoid FrozenInstanceError in `__setattr__ ` method of the `Env` class which is an equinox Module

- remove old mctx import


salvatomm (approved):
Run on my jax version and it worked fine, approved

Issue #251, opened 2025-09-03
ensure that `examples/envs/knapsack_demo.ipynb` works with `rollout()` function and `batch_size` arg


conorheins:

conorheins:
Closed with merge of #252

PR #252, opened 2025-09-03
Fix/251 update knapsack demo notebook

conorheins:
Addresses #251 
- replace numpy-controlled RNGs with JAX PRNGKeys
- upgrade use of Distribution object in notebook to latest
- fix typos / clarify description in first markdown cell

### DIFF
--- a/examples/envs/knapsack_demo.ipynb
+++ b/examples/envs/knapsack_demo.ipynb
@@ -15,7 +15,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": 1,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -37,7 +37,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 10,
+   "execution_count": 2,
    "metadata": {},
    "outputs": [
     {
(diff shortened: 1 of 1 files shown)


examples/envs/knapsack_demo.ipynb
```diff
@@ -126,9 +125,9 @@
      "text": [
       "A shapes [(2, 2), (2, 2), (2, 2), (2, 2), (2, 2), (2, 2)]\n",
       "B shapes [(2, 2, 2), (2, 2, 2), (2, 2, 2), (2, 2, 2), (2, 2, 2), (2, 2, 2, 2, 2, 2, 2)]\n",
-      "A normalized [True, True, True, True, True, True]\n",
-      "B normalized [True, True, True, True, True, True]\n",
-      "C normalized [True, True, True, True, True, True]\n"
+      "A normalized [np.True_, np.True_, np.True_, np.True_, np.True_, np.True_]\n",
```
Arun-Niranjan: what's the significance of `np.True_` instead of just `True`?

conorheins: In NumPy 2.0 (per [NEP 51](https://numpy.org/neps/nep-0051-scalar-representation.html)), the repr of NumPy scalars changed to include the type. So a `numpy.bool_` now displays as `np.True_`/ `np.False_` rather than True/False, to distinguish NumPy scalars from Python scalars. There’s no behavioral change here—`np.isclose(...).all()` still returns a `numpy.bool_`. If we want plain Python booleans in the printout, we can wrap with `.item()` or `bool(...)`.

Arun-Niranjan (approved):

PR #255, opened 2025-09-04
Get jax learning tests green

Arun-Niranjan:
now running `pytest test` on the chore/notebook branch will give a green result.

however I did mark the `test_update_state_likelihood_multi_factor_some_factors_no_action_2` as skipped. This test is failing due to the following error:

```
FAILED test/test_learning_jax.py::TestLearningJax::test_update_state_likelihood_multi_factor_some_factors_no_action_2 - ValueError: List arity mismatch: 2 != 3; list: [1, 2].
```

however it's not clear to me whether we even need this test or not, given that `test_update_state_likelihood_multi_factor_some_factors_no_action` also exists and is now passing.

Also if we want I can update the github workflow to also run the unit tests on all PRs, rather than just ones on the main branch?

### DIFF
--- a/test/test_learning_jax.py
+++ b/test/test_learning_jax.py
@@ -6,6 +6,7 @@ __author__: Dimitrije Markovic, Conor Heins
 """
 
 import unittest
+import pytest
 
 import numpy as np
 import jax.numpy as jnp
@@ -161,7 +162,7 @@ class TestLearningJax(unittest.TestCase):
             q_prev_f = jnp.array([qs_prev[..., f].tolist()])
             belief_jax.append([q_f, q_prev_f])
 
-        pB_updated_jax, _ = update_pB_jax(pB_jax, belief_jax, action_jax, num_controls=num_controls, lr=l_rate)
+        pB_updated_jax, _ = update_pB_jax(pB_jax, B, belief_jax, action_jax, num_controls=num_controls, lr=l_rate)
 
         for pB_np, pB_jax in zip(pB_updated_numpy, pB_updated_jax):
             self.assertTrue(pB_np.shape == pB_jax.shape)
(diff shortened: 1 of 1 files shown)


test/test_learning_jax.py
```diff
@@ -333,6 +334,7 @@ def test_update_state_likelihood_multi_factor_some_factors_no_action(self):
             self.assertTrue(pB_np.shape == pB_jax.shape)
             self.assertTrue(np.allclose(pB_np, pB_jax))
 
+    @pytest.mark.skip("currently failing with error: ValueError: List arity mismatch: 2 != 3; list: [1, 2].")
```
Arun-Niranjan, line 337: do we even need this test at all?

Arun-Niranjan:
It's a good question - I wanted it fixed because:
- I want to fix the model inversion notebook
- to do so I need to fix evolve trials
- to do that with confidence I want unit tests that pass
- so I thought I'd run the tests and I saw this was broken, so figured I'd get it green

so it's the dependency of fixing the model inversion notebook on having green tests which is why I've put it in this branch. Reminds me of the "[Hal's lightbulb](https://www.youtube.com/watch?v=5W4NFcamRhM)" episode of Malcolm in the middle 😂

Arun-Niranjan:
we are currently in the unfortunate position that the notebooks are currently acting like our tests on this branch, so I'm inclined to say that where adding/fixing tests to help us tidy up the existing notebooks specifically is reasonable. Thanks for the discussion!

test/test_learning_jax.py
conorheins: Just looked over this -  yes unfortunately there are several use cases that come to mind, that require this sort of update to be possible. The unit test directly above this one `test_update_state_likelihood_multi_factor_some_factors_no_action()` seems to be a kludge by @toonvdm that I assume he implemented when he found the naive way for implementing this type of learning rule doesn't work (the`test_update_state_likelihood_multi_factor_some_factors_no_action_2` that gives your list parity mismatch error).

I will take a deeper look and see what's going on here, after a 10 minute glance I'm having a hard time parsing what the issue is

conorheins:
Also sorry, just saw this comment

> Also if we want I can update the github workflow to also run the unit tests on all PRs, rather than just ones on the main branch?

Yes - that is a good idea IMO, go for it

conorheins (approved):
Approving now in line with @Arun-Niranjan 's previous comment and then will open another PR to unskip the skipped test with the fix to `learning.py`

PR #258, opened 2025-09-08
Enable running tests on chore/notebook pull requests

Arun-Niranjan:
Github actions have to be added to the master branch to be picked up, even if they only apply to other branches

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -23,6 +23,6 @@ jobs:
       - name: Install dependencies
         run: |
           uv sync
-      - name: Analysing the code with pylint
+      - name: Run tests
         run: |
           uv run pytest test


.github/workflows/test.yaml
```diff
@@ -0,0 +1,28 @@
...
+    strategy:
+      matrix:
+        python-version: ["3.12"]
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install uv
```
Arun-Niranjan, line 21: I presume we want to move to uv given we have a pyproject.toml file on the chore/notebook_cleanup branch? although I didn't see a uv.lock file

conorheins:
Looks good, but just wondering if we should go straight to approving your other (currently, draft) PR #257 , rather than going through this branch-specific workflow file first? Or was there a reason you wanted to do it this way

conorheins:
> https://github.com/infer-actively/pymdp/pull/257 is in draft because once I land this, if I make a small change to that one it should trigger the unit tests - it's my canary so to speak

Nice, understood. Will merge this in then

PR #272, opened 2025-09-11
fixed bug to select positions by dependency list indices

riddhipits:
Fixed `select_probs` function in `pymdp/envs/env.py` by changing `tuple(_float_to_int_index(p) for i, p in enumerate(positions) if i in dependency_list)` to `tuple(_float_to_int_index(positions[i]) for i in dependency_list)`. The original code was selecting position values in wrong order, and now it selects positions by dependency list indices in the correct order. I checked if the other notebooks in `examples/envs/` still run correctly with this change and they do.


conorheins (approved):
Good catch! Approved & merging

Issue #274, opened 2025-09-19
Inconsistent logic in updating A and B in `infer_parameters()` when `self.inference_mode = "ovf"`


conorheins:
While refactoring the `examples/envs/learning/learning_gridworld.ipynb` notebook as part of #208 and #210 I ran into the following issue when trying to do learning with `self.inference_mode = "ovf"`:

If you're at the first timestep, `self.learn_B=True`and `self.inference_algo = "ovf"`, then the inputs to the call to `smoothing_ovf` in `pymdp/agent.py:361` will not be consistent with their expected shape, which is 

```python
# beliefs_A (B, T, Ns), self.B (B, Ns, Ns, Nu), actions (B, T, Nu)
smoothed_marginals_and_joints = vmap(inference.smoothing_ovf)(beliefs_A, self.B, actions)
```

At the first timestep, `beliefs_A` (within the call of `agent.infer_parameters()`, which is the first argument to it `qs` in the call of `pymdp/envs/rollout.py:74`) will be of shape `(B, 1, Ns_f)`. This will lead to mismatched axes for the `lax.scan` call in the backward smoother pass of `smoothing_ovf` of `joint_dist_factor` of `pymdp/inference.py:110`. This is because `qs_last` will be of shape `(Ns_f,)` and `qs_filter` will be of shape `(0, Ns_f)` and `actions` will be of shape `(1, Nc_f)`.

This leads to the classic ValueError for `lax.scan`

```python
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File ".../pymdp/pymdp/inference.py", line 110, in joint_dist_factor
    _, seq_qs = lax.scan(
                ^^^^^^^^^
  File ".../pymdp/.venv311/lib/python3.11/site-packages/jax/_src/traceback_util.py", line 180, in reraise_with_filtered_traceback
    return fun(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^
  File ".../pymdp/.venv311/lib/python3.11/site-packages/jax/_src/lax/control_flow/loops.py", line 247, in scan
    raise ValueError(msg.format(', '.join(str(x.shape[0]) for x in xs_flat)))
ValueError: scan got values with different leading axis sizes: 0, 1.
````

[...]

Issue #211, opened 2025-08-05
Add `learning_mode` option to enable either online or offline learning


conorheins:
Suggested by @lancelotdacosta: we should add a `learning_mode` argument that either uses online or offline learning. This could be either a property of the `Agent` class or a functional argument to the `rollout()` function. 

The current version of the `rollout` function assumes if the agent can learn parameters (i.e. either `learn_A` or `learn_B` are True), then the learning is performed online. This assumption should be relaxed.

Let's use the `examples/learning/learning_gridworld.ipynb` notebook as a test case for when learning_mode = "offline", and thereby in the cross-fire of implementing this feature address #208  #210

Issue #210, opened 2025-08-05
Update `examples/learning/learning_gridworld.ipynb` with latest version of the `rollout` function


conorheins:
This notebook needs to be updated and currently doesn't run due to the following error when running the loop over blocks and agents

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[5], line 9
      7 for i, agent in enumerate(agents):
      8     key, _key = jr.split(key)
----> 9     _, grid_world = grid_world.reset(_key)
     11     key, _key = jr.split(key)
     12     last, info = jit(rollout, static_argnums=[3,] )(_key, agent, grid_world, num_timesteps)

    [... skipping hidden 3 frame]

File ~/.../pymdp_dev_env/lib/python3.11/site-packages/jax/_src/api.py:1246, in _mapped_axis_size(fn, tree, vals, dims, name)
   1244   else:
   1245     msg.append(f"  * some axes ({ct} of them) had size {sz}, e.g. axis {ax} of {ex};\n")
-> 1246 raise ValueError(''.join(msg)[:-2])

ValueError: vmap got inconsistent sizes for array axes to be mapped:
  * most axes (5 of them) had size 20, e.g. axis 0 of argument self.params['A'][0] of type float32[20,49,49];
  * one axis had size 2: axis 0 of argument key of type uint32[2]
  *
```

In addition to updating this notebook with the latest `Env` API (which should address the grid_world.reset(_key) line), I propose we also use the `rollout` function defined in `rollout.py`, rather than that used in-line in the notebook. This is related to #208 and Issue #

Issue #208, opened 2025-08-04
Fix `examples/learning/learning_gridworld.ipynb` notebook with latest `Env` API


conorheins:
While fixing the `examples/learning/learning_gridworld.ipynb` in a branch to address #204 , I ran into this error while running the active inference loop over blocks and agents

Jupyter notebook cell
```
pB0 = 1e4 * B[0] + 1e-4
num_timesteps = 50
num_blocks = 40
key = jr.PRNGKey(0)
divs = {i : [] for i in range(len(agents))}
for block in range(num_blocks):
    for i, agent in enumerate(agents):
        key, _key = jr.split(key)
        _, grid_world = grid_world.reset(_key)

        key, _key = jr.split(key)
        last, info = jit(rollout, static_argnums=[3,] )(_key, agent, grid_world, num_timesteps)
        grid_world = last['env']

        beliefs = info['qs']
        actions = info['actions']
        outcomes = info['observations']

        agents[i] = agent.infer_parameters(beliefs, outcomes, actions)
        divs[i].append(kl_div_dirichlet(agents[i].pB[0], pB0).sum(-1).mean(-1))
```

Error

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[6], line 9
      7 for i, agent in enumerate(agents):
      8     key, _key = jr.split(key)
----> 9     _, grid_world = grid_world.reset(_key)
     11     key, _key = jr.split(key)
     12     last, info = jit(rollout, static_argnums=[3,] )(_key, agent, grid_world, num_timesteps)

    [... skipping hidden 3 frame]

File ~/.../pymdp_dev_env/lib/python3.11/site-packages/jax/_src/api.py:1246, in _mapped_axis_size(fn, tree, vals, dims, name)
   1244   else:
   1245     msg.append(f"  * some axes ({ct} of them) had size {sz}, e.g. axis {ax} of {ex};\n")
-> 1246 raise ValueError(''.join(msg)[:-2])

ValueError: vmap got inconsistent sizes for array axes to be mapped:
  * most axes (5 of them) had size 20, e.g. axis 0 of argument self.params['A'][0] of type float32[20,49,49];
  * one axis had size 2: axis 0 of argument key of type uint32[2]
```

PR #280, opened 2025-09-26
Fix/update learning gridworld notebook 208 210

conorheins:
This PR addresses #208 #210 #211 #274 by updating the `examples/learning_gridworld.ipynb` notebook. 

In order to refactor this notebook to use the new `Env` and `rollout` functions, the following changes were made in the crossfire:
- add an "online" vs "offline" learning mode toggle to `Agent` class (e.g., `agent(..., learning_mode="online")`, which determines when parameters are updated in `rollout()`. "online" learning means that parameters are updated for every call of `infer_and_plan()` and thus at every timestep of the scanned rollout. "offline" learning means that parameters are updated after the scanned rollout, and before the return, using the history of (possibly-smoothed) posteriors, observations, and actions.
- adds a `GridWorld` environment to the `envs/` folder and a corresponding unit test to ensure parity with the original NumPY `GridWorldEnv` class
- refactor other notebooks that use the `GridWorld` class (`inductive_inference_gridworld.ipynb`) to use the new JAX-backend env
- clean up the logic behind how `smoothing_ovf` is called inside the `infer_parameters()` method of `Agent`. An additional `update_B` flag is computed internally and used to conditionally decide whether to compute the smoothing distribution or not, depending on whether at least one action has been taken (and thus there are two timesteps worth of posteriors to use for smoothing). This addresses #274


pymdp/agent.py
```diff
@@ -91,6 +92,8 @@ class Agent(Module):
     sampling_mode: str = field(static=True)
     # fpi, vmp, mmp, ovf
     inference_algo: str = field(static=True)
+    # whether to perform leraning online or offline (options: "online", "offline")
```
Arun-Niranjan:
```suggestion
    # whether to perform learning online or offline (options: "online", "offline")
```

pymdp/agent.py
```diff
@@ -131,6 +134,7 @@ def __init__(
         inference_algo="fpi",
         num_iter=16,
         batch_size=1,
+        learning_mode="online", # TODO: or should this be an argument to `self.infer_parameters()` or even `env/rollout.py:rollout()`
```
Arun-Niranjan, line 137: I think this question is cropping up a lot. perhaps eventually `rollout` needs to be a method on the `Agent` class itself?

I would be a bit concerned that we have some settings on the `Agent`, some on functions that are applied to an `Agent` and they could potentially conflict - this could lead to plenty of footguns for users.

pymdp/agent.py
```diff
@@ -356,17 +361,74 @@ def _get_num_states_from_B(self, B, B_dependencies):
...
+        seq_beliefs = beliefs_A if beliefs_B is None else beliefs_B  # (list over factors) each (B, T, Ns_f)
+
+        # Normalize action shape to include time if needed
+        if actions is not None and actions.ndim == 2:
+            # (B, Nu) -> (B, 1, Nu) so we can slice to T-1 below
+            actions = jnp.expand_dims(actions, 1)
+        
+        # Infer sequence length (T) from beliefs provided for B learning
+        # If learn_B is False we don't need a time sequence, but computing T is cheap and safe
+        T = seq_beliefs[0].shape[1]  # each factor has (B, T, Ns_f)
```
Arun-Niranjan, line 379: Is this different to the `policy_len` parameter on the agent?

pymdp/agent.py
```diff
@@ -356,17 +361,74 @@ def _get_num_states_from_B(self, B, B_dependencies):
...
+
+        # Make actions time-conformant: always slice to T-1 along time
+        if actions is not None and actions.ndim == 3:
+            actions_Tm1 = actions[:, :max(T - 1, 0), :]    # (B, max(T-1, 0), Nu)
+        else:
+            actions_Tm1 = actions  # None or already time-matched
+        
+        # A handy predicate: can we (meaningfully) update B now?
+        # We need at least one transition and all actions to be non-negative (sentinel -1 = invalid).
```
Arun-Niranjan, line 388: for my understanding - when would an Agent be able to select an invalid action?

Arun-Niranjan: ah I've inferred this might be due to the environment i.e. the agent wants to move left but it hits a wall. If that's the case, wouldn't the agent need to update it's B-matrix accordingly?

test/test_rollout_function.py
```diff
@@ -0,0 +1,288 @@
...
+import jax.random as jr
+import jax.tree_util as jtu
+
+from pymdp.legacy import utils
+from pymdp.agent import Agent
+from pymdp.envs.env import Env
+from pymdp.envs.rollout import rollout, default_policy_search
+
+
+class TestRolloutFunction(unittest.TestCase):
```
Arun-Niranjan, line 19: great effort on these unit tests. One question I have is that rollout returns the env after rollout, but I don't think I've ever seen that returned environment used after rollout is complete.

Perhaps that needs a unit test as well, or removing from the rollout outputs?

pymdp/agent.py
Arun-Niranjan: I think for now our strategy should be to:
- add complexity where we need it to the Agent
- establish where inside the agent data vs settings are being used (this may be algorithm specific)
- carve the agent at the joints to make it more modular and testable

pymdp/agent.py
dimarkov: Yes, T here is the length of the observed sequence

dimarkov (approved):

pymdp/agent.py
conorheins: So `policy_len` is the planning horizon, which is different from `T`, which as @dimarkov said is the length of the observed sequence (how many observations have been seen). As an example, you could have an episode that lasts 10 timesteps but you're only planning with a horizon of 2 steps ahead, so `T = 10, policy_len = 2`.

test/test_rollout_function.py
conorheins: Good point, will add a test for the `env` after it's been returned from the `rollout` function

pymdp/agent.py
conorheins: So at the first timestep of an active inference episode, we usually set the actions to -1, because the agent hasn't had the chance to perform a single step of (inference + planning + action selection). AFAIK that is the only case where we would expect action to be set to -1

pymdp/agent.py
conorheins: agreed, think this should be rolled into the API convo (relevant to both `Agent` and `Env`) we have after merging the `chore/notebook_cleanup` branch

test/test_rollout_function.py
conorheins: Actually I forgot that I already started a branch two weeks ago [here](https://github.com/infer-actively/pymdp/tree/tests/extend_rollout_function_testing) where I am extending the rollout tests. In the interest of keeping this PR focused on the online/offline learning toggle and its relationship to notebooks, I will put that `env`-check in that branch and PR it separately.

PR #281, opened 2025-10-04
Fix evolve_trials

Arun-Niranjan:
This change modifies `evolve_trials` to be compatible with `rollout` output and adds a unit test.


pyproject.toml
```diff
@@ -33,6 +33,7 @@ dependencies = [
     'mctx>=0.0.5',
     'networkx>=3.3',
     'pytest>=6.2.1',
+    "numpyro>=0.19.0",
```
Arun-Niranjan, line 36: this isn't actually used in `evolve_trials`, but it is in `aif_likelihood`, which we'll fix in a separate PR

Issue #239, opened 2025-08-29
Add tests for jax demos


conorheins:
Mentioned in the description for #238 , we should add tests for the new JAX-backend demos in the examples/ folder of pymdp.

This issue is also related to / addressed by #293

PR #283, opened 2025-10-09
feat: add setter in Distribution that validates tensor shape match

coopwilliams:
It's common to edit or even replace a tensor in the model, after declaring it using the Distribution API. Since the tensor shape is a function of multiple fields in the model, it's unlikely that changing its shape directly would result in a valid model. Generally it should be much easier to make those modifications within a single block of code, i.e., the model description. Changing the shape of the tensor should be an invalid action, since it's likely to be user error.

This PR adds a setter in the Distribution class that prevents the user from replacing a tensor with a differently shaped one. Doing so throws a clear error message.

All tests passing here.

Addresses https://github.com/infer-actively/pymdp/issues/288

### DIFF
--- a/pymdp/distribution.py
+++ b/pymdp/distribution.py
@@ -18,28 +18,39 @@ class Distribution:
         }
 
         if data is not None:
-            self.data = data
+            self._data = data
         else:
             shape = []
             for v in event.values():
                 shape.append(len(v))
             for v in batch.values():
                 shape.append(len(v))
-            self.data = np.zeros(shape)
+            self._data = np.zeros(shape)
+
+    @property
+    def data(self):
+        return self._data
+
+    @data.setter
+    def data(self, value):
+        if hasattr(self, '_data') and self._data is not None:
+            if self._data.shape != value.shape:
+                raise ValueError(f"When setting the tensor directly, the new shape {value.shape} must match the existing tensor shape {self._data.shape}")
+        self._data = value
 
     def get(self, batch=None, event=None):
         event_slices = self._get_slices(event, self.event_indices, self.event)
         batch_slices = self._get_slices(batch, self.batch_indices, self.batch)
 
         slices = event_slices + batch_slices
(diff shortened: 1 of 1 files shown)


pymdp/distribution.py
```diff
@@ -18,28 +18,39 @@ def __init__(self, event: dict, batch: dict = {}, data: np.ndarray = None):
...
+    @property
+    def data(self):
+        return self._data
+
+    @data.setter
+    def data(self, value):
+        if hasattr(self, '_data') and self._data is not None:
+            if self._data.shape != value.shape:
+                raise ValueError(f"When setting the tensor directly, the new shape {value.shape} must match the existing tensor shape {self._data.shape}")
+        self._data = value
```
Arun-Niranjan, line 39: my understanding is that by using the property decorator, you're able to change the field name internally without this being a breaking change. Nice one!

conorheins:
> @conorheins I believe this extra validation should be fine with the JAX implementation as we never create a `Distribution` outside of `Agent` initialisation.
> 
> wanted to double check my understanding with you, otherwise this looks good to me

Yes I think this should be fine / is a good extra check to have, to avoid inconsistent shapes of likelihoods already at the level of the Distribution.

I'll check that it doesn't break /give unexpected outputs for any of the existing `examples/` that rely on Distribution (or someone else can, if they have time -- just verify those example notebooks give expected outputs in a comment here), before approving/merging. Due to #239 I still have been manually checking notebook outputs don't break with every PR that affects functionality in any of the notebooks

Arun-Niranjan:
> I'll check that it doesn't break /give unexpected outputs for any of the existing examples/ that rely on Distribution (or someone else can, if they have time -- just verify those example notebooks give expected outputs in a comment here), before approving/merging

I can check this manually now, and will raise a ticket for checking the notebooks as part of the test suite using [nbval](https://pypi.org/project/nbval/)

Issue #295, opened 2025-10-22
Add unit test for jittability/autodiffability of `Agent` construction


conorheins:
In model inversion workflows with [pybefit](https://github.com/dimarkov/pybefit) and [numpyro](https://github.com/pyro-ppl/numpyro), a user will often have to write a custom parameterization function `transform()` function that takes values of a latent variable (e.g., samples) and uses them to parameterize a pymdp agent with those samples, e.g.,

```python
def transform(z):
    A, B, C, D, E = parameterize_model_from_samples(z) # some code here that parameterizes the probability tensors using samples
    return Agent(A, B, C, D, E)
```

Since this function will be called iteratively in the course of a NUTS / MCMC loop or a stochastic variational inference loop, the `Agent` constructor (`__init__`) needs to be autodifferentiable and jittable. We should have a unit test that ensures this, e.g.

```python

def test_agent_construction_jittability(...)
     def construct_agent(params):
            return Agent(**params)
     jit(construct_agent)(params) # make sure this runs
```

and likewise one that returns well-defined/non-NaNed gradients e.g. ensuring `~jnp.any(jnp.isnan(jax.grad(construct_agent)(params)))`

conorheins:
The ideal unit test here should also test the differentiability of the one-step active inference process for a set of plausible different configurations of the `Agent` class (different settings of learning and inference algorithms), to make sure all those sub-processes are also autodifferentiable.

Issue #287, opened 2025-10-15
refactor `model_inversion.ipynb` notebook to use only functions from `pybefit` package


conorheins:
This will require adding an optional `pybefit` dependency (via a dependency group in the `pyproject.toml`) to pymdp and removing the now deprecated `likelihoods.py` file from pymdp

Issue #242, opened 2025-09-03
Update examples to use latest `batch_size` parameter and `rollout` function - /model-fitting


Arun-Niranjan:
- [ ] Update the model_inversion.ipynb notebook with the batch_size argument for Agent constructor and new rollout function -- if applicable

Arun-Niranjan:
blocked on https://github.com/infer-actively/pymdp/issues/241

PR #296, opened 2025-10-27
Cleanup/update model inversion notebook

### LINKED ISSUE #201: Update example notebooks to handle batch size parameter
based on https://github.com/infer-actively/pymdp/issues/173 and github.com/infer-actively/pymdp/pull/186

conorheins:
To address #242 (and thus close #201), #287, #295

Updates the `examples/model_fitting` notebook with the updates to how the `batch_size` argument is passed to the `Agent` constructor, among other minor API updates (like the structure of the return of `self.reset()` method of `pymdp.envs.Env`).

### Summary of changes
- add an optional `modelfit` dependency group to the `pyproject.toml`, which adds the `pybefit` package as a dependency. For the time being (until these changes are merged to `v1.0.0_alpha` and then to `master`), this is pointing to a branch of `pybefit` called `fix/pymdp-compat` on the `conorheins` fork.
- remove deprecated `model_inversion.ipynb` notebook which assumed deprecated APIs of both `pymdp` and `pybefit`, and replaced it with a fresh new notebook called `fitting_with_pybefit.ipynb`, which fully depends on `pybefit`'s custom likelihood functions for `pymdp`. See discussion in thread of #281 and #287 for context / motivation for this decision. This notebook is currently running without errors or NaNs, but the parameter recovery plots are not good (no correlation between ground truth parameter values and the inferred ones), but this is likely an issue with the model identifiability rather than the functioninig of `pybefit`. It uses the most up-to-date version of the TMaze task. The `pybefit` package itself has an `examples/pymdp/` folder with a functioning model inversion example (see [here](https://github.com/conorheins/pybefit/blob/fix/pymdp-compat/examples/pymdp/simulating_and_fitting_agency_task.ipynb), based on a real [agency experiment](https://osf.io/preprints/osf/ckfrt) from computational psychology (by @riddhipits and @dimarkov).
- adds unit tests for jittability of `__init__` of `Agent` (needed for compatibility with `pybefit`) and for autodifferentiability of: building an agent and running a single timestep of active inference , addressing #295. Also refactored the normalization checking function in `utils.py` to use an equinox `error_if` function which allows graceful handling of runtime errors in JIT context.


test/test_agent_jax.py
```diff
@@ -490,6 +494,205 @@ def test_agent_with_A_learning_requires_pA(self):
...
+            return agent
+        
+        # write a unit test that the Agent constructor is jittable
+        jitted_constructor = jit(construct_agent)
+        agent_instance = jitted_constructor(
+            A_batched, B_batched, pA_batched, pB_batched, D_batched
+        )
+        self.assertIsInstance(agent_instance, Agent)
+
+    def test_valid_gradients_one_step_ahead(self):
```
Arun-Niranjan, line 557: At the moment I believe this test will fail at the first exception, which means we would lose coverage on the subsequent ones. It probably makes sense to break this test out into parameterized sub tests - one for each inference algorithm, with the added benefit of being able to run each version of the test in parallel.
https://docs.pytest.org/en/stable/how-to/parametrize.html
But that's something I will take a look at separately

Arun-Niranjan (commented):
I can't check the notebook from my phone, but the rest of the changes + the tests look good to me.

test/test_agent_jax.py
conorheins: Good point - i was being lazy and just enumerated the cases sequentially, but you're right we should absolutely parameteriz pytests like this (there are others that deserve similar treatment throughout the test suite). I've done this before and it's pretty straightforward. thanks for offering to tackle it

PR #315, opened 2025-11-12
Add optimized infer_states methods & demo notebook; Add unit test.

nikolamilovic-ft:
- Added three optimized infer_states methods: **end-to-end padded**, **hybrid**, and **hybrid block**
- Core logic implemented in `algos.py`, `maths.py`, and `utils.py`.
- Added a demo notebook - `methods_test.ipynb` - for testing and comparing the optimized methods.
- Introduced a new unit test script for verifying output consistency between the optimized and original implementations.
- `agent.py` remains unchanged until a design decision is made on integrating the new methods and the `process_obs` function.


nikolamilovic-ft:
Thanks @conorheins, definitely agree with the suggestions! I have implemented the requested changes. Please take a look at those when you find time. It would be nice to include Miguel (@praesc) in this conversation since we should also discuss how these methods could be integrated in `agent.py` scipt when the time comes.

pymdp/algos.py
```diff
@@ -347,6 +347,122 @@ def run_online_filtering(A, B, obs, prior, A_dependencies, num_iter=1, tau=1.):
     qs = update_marginals(get_mmp_messages, obs, A, B, prior, A_dependencies, num_iter=num_iter, tau=tau)
     return qs 
 
+
+# Infer states hybrid & hybdrid block
```
conorheins:
```suggestion
# Infer states hybrid & hybrid block
```

conorheins (changes requested):
Thanks for addressing these comments, looks good. Just one final request -- can you rebase this now that #319 has landed?

Also can you move the `infer_states_optimization/` folder within `examples`  to `examples/advanced/`?

conorheins (approved):
Looks good, let's merge it in!

PR #294, opened 2025-10-21
Add automated tests for notebooks (local only)

Arun-Niranjan:
This change addresses https://github.com/infer-actively/pymdp/issues/293 -  adding the infrastructure required for automated testing of notebooks. this includes:
- the addition of a `test` dependency group that has the existing test dependencies like pytest. `nbval` is the main new dependency for notebook evaluation
- adds a `test_notebook.py` file that will check if notebooks exist under normal test command execution (e.g. `uv run pytest test`)
- adds a readme explaining how to run the notebook tests locally with `nbval`

eventually with a bit more tidy-up, we can run these tests in CI (either on every PR or a nightly build if they take too long)

Issue #293, opened 2025-10-21
Investigate way of automatically running notebooks in CI


Arun-Niranjan:
we currently have coverage of pymdp functionality split across unit tests and example notebooks.

even if unit tests pass, we still have to manually run all our example notebooks to check if a PR is breaking anything

it would be good to try out the following solutions:
1. [nbval](https://pypi.org/project/nbval/)
2. [pytest-notebook](https://github.com/chrisjsewell/pytest-notebook)
3. [nbconvert](https://nbconvert.readthedocs.io/en/latest/index.html)

and decide on one going forward that means we can run jupyter notebooks as part of the CI test suite

Arun-Niranjan:
I'll split into sub issues and get at least a test for the fast notebooks (the api one goes quick) running in CI

Issue #268, opened 2025-09-10
Add section in README.md for guidance on installing pygraphviz


conorheins:
I had some trouble installing pygraphviz when trying to run 

```bash
>>> uv venv
>>> source .venv/bin/activate
>>> uv sync --activate --extra nb
```

on MacOS (Apple Silicon M2 Pro). I found the following workaround using 

```bash
brew install graphviz
uv pip install \
  --config-settings=--global-option=build_ext \
  --config-settings=--global-option="-I$(brew --prefix graphviz)/include" \
  --config-settings=--global-option="-L$(brew --prefix graphviz)/lib" \
  pygraphviz==1.14
````

But this should be documented somewhere explicitly or otherwise incorporated into how we install the dependency, perhaps in the README.md

UPDATE: I also ran into this issue when trying to install dependencies from scratch on Ubuntu 24.04 LTS. I ran 

```bash
uv sync --group test --extra gpu
```

and got this error

```
Updated https://github.com/conorheins/pybefit.git (f20c94d30fbf938bf671ee863a552
Resolved 180 packages in 11.66s
warning: The package `jaxlib==0.8.1` does not have an extra named `cuda12`
warning: The package `jaxlib==0.6.2` does not have an extra named `cuda12`
      Built inferactively-pymdp @ file:///home/conor/Documents/.../
      Built pybefit @ git+https://github.com/conorheins/pybefit.git@f20c94d30fbf938b
  × Failed to build `pygraphviz==1.14`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `setuptools.build_meta.build_wheel` failed (exit status: 1)

      [stdout]
      running bdist_wheel
      running build
      running build_py
      creating build/lib.linux-x86_64-cpython-312/pygraphviz
      copying pygraphviz/testing.py ->
      build/lib.linux-x86_64-cpython-312/pygraphviz
      copying pygraphviz/graphviz.py ->
      build/lib.linux-x86_64-cpython-312/pygraphviz
      copying pygraphviz/agraph.py ->
      build/lib.linux-x86_64-cpython-312/pygraphviz
      copying pygraphviz/scraper.py ->
      build/lib.linux-x86_64-cpython-312/pygraphviz
      copying pygraphviz/__init__.py ->
      build/lib.linux-x86_64-cpython-312/pygraphviz
      creating build/lib.linux-x86_64-cpython-312/pygraphviz/tests
      copying pygraphviz/tests/test_string.py ->
[...]

PR #319, opened 2025-11-20
Add local automated notebook tests

Arun-Niranjan:
This change addresses https://github.com/infer-actively/pymdp/issues/293 - adding the infrastructure required for automated testing of notebooks. this includes:

- the addition of a test dependency group that has the existing test dependencies like pytest. [nbval](https://nbval.readthedocs.io/en/latest/) is the main new dependency for notebook evaluation.
⚠️ pytest has been moved out of the core dependency group. This means that to be able to run tests you *must* run `uv sync --group test` to get the correct dependencies.
- adds a readme explaining how to run the notebook tests locally with nbval

eventually with a bit more tidy-up, we can run these tests in CI (either on every PR or a nightly build if they take too long)

this PR is largely reproduced from https://github.com/infer-actively/pymdp/pull/294 to account for the merge from chore/notebook to v1 alpha branch

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -20,6 +20,10 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install uv
         uses: astral-sh/setup-uv@v6
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y graphviz-dev
       - name: Install dependencies
         run: |
           uv sync --group test


test/notebooks/test_notebooks.py
```diff
@@ -0,0 +1,134 @@
...
+@pytest.mark.slow
+@pytest.mark.parametrize("notebook", ADVANCED_NOTEBOOKS)
+def test_advanced_notebooks(notebook):
+    """Test advanced feature notebooks."""
+    assert notebook.exists(), f"Notebook not found: {notebook}"
+
+
+@pytest.mark.notebook
+@pytest.mark.slow
+@pytest.mark.skip("model fitting notebook is currently being migrated, see https://github.com/infer-actively/pymdp/issues/287")
```
conorheins: I think this pytest.mark.skip can be removed now, because the notebook is currently running (but the pytest.mark.slow is certainly warranted)

pyproject.toml
```diff
@@ -74,3 +83,19 @@ line-length = 88
...
+    "slow: slow-running tests (run on schedule)",
+    "experimental: experimental/potentially unstable tests",
+]
+filterwarnings = [
+    "ignore::DeprecationWarning",
+    "ignore::FutureWarning",
+]
+
+[tool.uv.sources]
+pybefit = { git = "https://github.com/conorheins/pybefit.git", rev = "fix/pymdp-compat" }
```
conorheins, line 98: I guess because you have this line now, we can change the `modelfit` dependency to just use `pybefit` now, right? If that's the case, can you change that before we merge this? 

so

```
modelfit = [
    "pybefit"
]
```

test/notebooks/README.md
```diff
@@ -0,0 +1,37 @@
...
+uv run pytest --nbval examples/api/
+
+# Test specific notebook
+uv run pytest --nbval examples/api/model_construction_tutorial.ipynb -v
+```
+
+:warning: For now, notebook tests are opt-in, i.e. running `uv run pytest test` will only check that notebooks exist.
+This tool is currently intended to help test notebooks locally, and eventually we will move them into CI if we can keep execution times low.
+
+The legacy notebooks are not tested.
```
conorheins, line 34: Note on this, in case this is different on your end:

When i run `pytest --nbval-lax examples/` it also runs the legacy notebooks. That's fine as far as I'm concerned but maybe you can revise this to reflect that too, or find a way to explicitly skip the legacy notebooks if that was your intention

pyproject.toml
```diff
@@ -32,7 +32,16 @@ dependencies = [
...
     'networkx>=3.3',
-    'pytest>=6.2.1',
+]
+
+[dependency-groups]
+test = [
+    "ipykernel>=7.1.0",
+    "jupyter>=1.1.1",
+    "nbval>=0.11.0",
+    "pybefit",
```
conorheins, line 43: in the same way you're adding pybefit as a dependency to the `test` dependencty group, to be able to run the model-fitting notebook, can you also add the `mediapy` and `pygraphviz` dependencies here? otherwise people won't be able to run the notebook tests without all cells failing on tmaze_demo.ipynb, which requires mediapy (and is a pretty foundational notebook among the new jax suite)

test/notebooks/README.md
Arun-Niranjan: I had misunderstood how the pytest nbval plugin hooks into pytest - turns out the test_notebook.py file was actually not used by nbval at all - so I've removed it for simplicity (checking whether a notebook exists is actually a pretty useless test.)

I've instead added a line to the pyproject.toml file to ignore the examples/legacy folder when running pytest, which definitely is picked up by nbval

pyproject.toml
```diff
@@ -74,3 +85,14 @@ line-length = 88
 [tool.ruff]
 line-length = 88
 lint.ignore = ["E731", "E741", "F401", "E402", "F811"]
+
+[tool.pytest.ini_options]
+minversion = "6.2.1"
+addopts = "--ignore=examples/legacy"
```
Arun-Niranjan, line 91: this line means that `examples/legacy` is successfully ignored by pytest by default

.github/workflows/test.yaml
```diff
@@ -20,9 +20,13 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install uv
         uses: astral-sh/setup-uv@v6
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y graphviz-dev
```
Arun-Niranjan, line 26: this is required now we have pygraphviz as a test dependency

conorheins: thanks for catching that -- also for people running locally with the `test` dep group we're gonna have to deal with this somehow: https://github.com/infer-actively/pymdp/issues/268, but let's leave that to a future PR

Arun-Niranjan:
> Looks great thanks for addressing all those comments -- one final request before we merge - can you just squash the commits together into one ? I'm trying to keep the commit # as small as possible in advance of the v1.0.0_alpha --> master PR

yep there's a "squash and merge" button in github which does this for us 😄

PR #321, opened 2025-11-21
Add automated test for api notebooks to CI

Arun-Niranjan:
Addresses https://github.com/infer-actively/pymdp/issues/322

This change adds running the `examples/api` notebooks to our CI. This is done separately to the unit tests because incoming speed optimisations for running unit tests vs notebook tests will likely be very different 

There is only a single notebook and the test runs pretty [quickly](https://github.com/infer-actively/pymdp/actions/runs/19574811497/job/56057194663?pr=321) (adds ~15s in total to CI times )

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -32,4 +32,4 @@ jobs:
           uv run pytest test
       - name: Running notebook tests
         run: |
-          uv run pytest --nbval examples/api/
+          uv run pytest --nbval-lax examples/api/


.github/workflows/test.yaml
```diff
@@ -30,3 +30,6 @@ jobs:
       - name: Running unit tests
         run: |
           uv run pytest test
+      - name: Running notebook tests
+        run: |
+          uv run pytest --nbval examples/api/
```
Arun-Niranjan: @conorheins happy to add other notebooks to this, but to start with I want to keep the execution times low (the default github actions runner is CPU only).

once we also land https://github.com/infer-actively/pymdp/pull/323 we will have a bit more budget for running the other notebooks

conorheins (commented):
Cool this is great -- and yes agree with keeping it to just examples/api to start. Did you want to keep it milder to `--nbval-lax` to start with, or do you think we should already keep cell-by-cell parity with the notebooks? I know at least one of the notebooks in `examples/api/` will be changing soon (due to an upcoming feature of @tverbele ) so perhaps we can keep it to just `--nbval-lax` for this one for now? Doesn't matter much at the moment though, we can always change it if it causes issues.

Also can you please rebase this and trigger the CI again, because there is now another notebook in examples/advanced/ that is on v1.0.0_alpha with the landing of #315. I think it should be fine though, shouldn't mess with any of the notebook testing infra

Issue #320, opened 2025-11-21
Speed up the unit tests


Arun-Niranjan:
currently unit tests take ~6m 30s to run on my local machine, and ~4 minutes on github actions. It would be good to get this down

there are likely some speedups we can introduce:
- run tests in parallel using pytest-xdist
- simplify some of the tests to not use as large models
- investigate some JAX settings (caching, memory allocation etc.)

Care should be taken to ensure we keep good test coverage and prioritise correctness over speed - as pymdp is a package implementing some tricky maths, unit tests are one of our biggest assets in being confident in our implementations

PR #323, opened 2025-11-22
Use pytest-xdist to run tests in parallel

Arun-Niranjan:
Addresses https://github.com/infer-actively/pymdp/issues/320

This change: 
- uses the `pytest-xdist` plugin to run tests concurrently, with 1 worker by default.
- Sets the `XLA_PYTHON_CLIENT_ALLOCATOR` and `XLA_PYTHON_CLIENT_MEM_FRACTION` [environment variables](https://docs.jax.dev/en/latest/gpu_memory_allocation.html) to sensible values if they are not already set
- Sets the `JAX_COMPILATION_CACHE_DIR` [environment variable](https://docs.jax.dev/en/latest/persistent_compilation_cache.html) if it isn't already

On my local machine, using 2 workers cut the test suite execution time down from ~6m 30s to ~4m

On CI (where 2 workers are used explicitly), we can see that execution time has dropped from [~5m](https://github.com/infer-actively/pymdp/pull/321/checks) to [~3m 30s](https://github.com/infer-actively/pymdp/actions/runs/19598843171/job/56127342688?pr=323)

The xdist settings defaults in pyproject.toml can be overridden when running tests locally, e.g. with 
`uv run pytest test -n auto --dist loadscope`

### DIFF
--- a/test/test_infer_states_optimized.py
+++ b/test/test_infer_states_optimized.py
@@ -65,7 +65,7 @@ class TestInferStatesComparison(unittest.TestCase):
         parameter_sets = [
             (5, 5, 5, 5, 'uniform', 'low'),
             (10, 10, 10, 10, 'uniform', 'medium'),
-            # (25, 25, 25, 25, 'uniform', 'high'),
+            (25, 25, 25, 25, 'uniform', 'high'),
             # (125, 125, 125, 125, 'uniform', 'extreme'),  # Uncomment to include extreme cases
         ]
 
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -19,7 +19,7 @@ def pytest_configure(config):
 
     # Limit memory fraction per device (to prevent OOM kills)
     if 'XLA_PYTHON_CLIENT_MEM_FRACTION' not in os.environ:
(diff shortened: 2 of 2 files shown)


Arun-Niranjan:
@conorheins after a rebase I am seeing memory leaks when I run the tests locally even with two workers, I've put the PR in draft until I can debug and fix

test/test_infer_states_optimized.py
```diff
@@ -65,7 +65,7 @@ def setUpClass(cls):
         parameter_sets = [
             (5, 5, 5, 5, 'uniform', 'low'),
             (10, 10, 10, 10, 'uniform', 'medium'),
-            (25, 25, 25, 25, 'uniform', 'high'),
+            # (25, 25, 25, 25, 'uniform', 'high'),
```
Arun-Niranjan: I believe it's this level of model complexity that was causing unit tests to fail on my local machine and on CI - needing too much memory

I think long term we should have separate benchmark tests that live outside the unit tests and run on a different cadence

Arun-Niranjan: addressed this with a bit more tuning of the memory allocation per worker

Issue #277, opened 2025-09-22
rollout function is incompatible with onehot_obs Agent?


Arun-Niranjan:
With an initial setup like so:

```python
from jax import random as jr
from pymdp.agent import Agent
from pymdp.envs import TMaze, rollout

key = jr.PRNGKey(0)
batch_size=1
T = 5
env = TMaze(
    batch_size=batch_size,
)
A, A_dependencies = env.generate_A()
B, B_dependencies = env.generate_B()
agent = Agent(
    A=A,
    B=B,
    A_dependencies=A_dependencies,
    B_dependencies=B_dependencies,
    batch_size=batch_size,
    policy_len=T,
    onehot_obs=True,
)
_, info, _ = rollout(agent, env, num_timesteps=T, rng_key=key)
```
`rollout` raises an exception when trying to call `agent.infer_states(...)`. 

This does not occur when setting `onehot_obs=False` in the `Agent` constructor.  We should either:
1. fix rollout to handle one hot encoded observations (preferable)
2. remove the flag from the agent constructor

if we keep support for one hot encoded observations, we should add unit tests to cover this option

conorheins:
Good catch, and more reason for us to get unit tests for rollout function integrated sooner rather than later

tverbele:
The jax-compatible way of doing the above is to allocate for some maximum, i.e. have a state or obs dimension of K, initialize your agent only using k < K. Then over time you can start using extra dimensions and parameter slices as data comes in. The other option is to completely re-create your Agent object with expanded parameters, but this implies also having to re-jit all the functions for the new sizes.

Arun-Niranjan:
I think for now the good middle ground to strike before a v1 release is either a K-vector *or* an individual value for observations, and support structure learning in the observation space as a follow on iteration

> A subtlely different but related case is hierarchical models where one agents observations are the hidden state posteriors of another agent (a lower level POMDP).

This is the 'killer app' in my mind, anything we can do to make the developer experience easier to achieve this will pay dividends

PR #326, opened 2025-11-29
Add support for categorical observations in Agent

Arun-Niranjan:
Addresses https://github.com/infer-actively/pymdp/issues/327

This change makes it clearer how to use a belief distribution for observations as opposed to a discrete observation.

- ⚠️ Breaking Change `onehot_obs` -> `categorical_obs` in `Agent` API and functions such as `infer_states`
- passing `categorical_obs` down the stack into relevant functions
- new unit tests covering agent and function operation when supplied with categorical observations

*Note* This change does not make the `rollout` function compatible with observation vectors as per https://github.com/infer-actively/pymdp/issues/277, that is handled in https://github.com/infer-actively/pymdp/pull/328

### DIFF
--- a/pymdp/agent.py
+++ b/pymdp/agent.py
@@ -25,7 +25,7 @@ class Agent(Module):
 
     The basic usage is as follows:
 
-    >>> my_agent = Agent(A = A, B = C, <more_params>)
+    >>> my_agent = Agent(A = A, B = B, C = C, <more_params>)
     >>> observation = env.step(initial_action)
     >>> qs = my_agent.infer_states(observation)
     >>> q_pi, G = my_agent.infer_policies(qs)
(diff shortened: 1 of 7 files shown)


Arun-Niranjan:
> looks good to me, but is there a reason to use `categorical_obs` in one method but then `distr_obs` in another?

thought I'd keep it the same as that seemed like a more internal method - one thing I'd like to do as part of the API discussions is "hide" internal methods to make it easier to refactor those without having to worry about breaking changes for users.

pymdp/agent.py
```diff
@@ -472,31 +492,71 @@ def use_filtered_beliefs(_):
 
         return agent
 
-    def infer_states(self, observations, empirical_prior, *, past_actions=None, qs_hist=None, mask=None, onehot_obs=False):
+    def infer_states(self, observations, empirical_prior, *, past_actions=None, qs_hist=None, mask=None, categorical_obs=False):
```
conorheins, line 495: Not a full review here, but just wanted to add a comment for future notice -- I am planning on refactoring observation processing into a separate method `process_obs()` so that it can be separated from the actual inference execution, following the discussions surrounding the #315 optimized inference methods and how to best integrate them into `Agent`. I created a small issue detailing this here: https://github.com/infer-actively/pymdp/issues/330

So this doesn't conflict with that, but just a heads up that these docstrings for  `infer_states()` may become deprecated or have to be repurposed once that `process_obs()` method gets added.

Issue #279, opened 2025-09-26
a reset env gets overriden when combined with `rollout`


conorheins:
Let's imagine a case where I have an environment instance and then i reset it to some desired initial condition,

```python
_, reset_env = env.reset(jr.split(key, batch_size), init_states_per_batch)
````

And then i want to use that reset environment in the rollout function, using

```python
last, info, final_env = rollout(agent, reset_env, num_timesteps, key, ...)
````

The initial state that was fixed from the reset (`init_states_per_batch`) will be ignored because of [line 182 of `rollout.py`](https://github.com/infer-actively/pymdp/blob/55fb84cd9f889923cb5a6515b8aa12273702dd0c/pymdp/envs/rollout.py#L182), where the environment is reset without a `state` argument, meaning that `env.params['D']` is used to sample an initial state

Some options I foresee to solve this are:

1) when doing a reset with an externally provided initial state, also reset `env.params["D"]` using something like

```python 
import equinox as eqx
from jax import tree_util as jt
from jax import nn as jnn
onehotted_states = jtu.tree_map(lambda s_f, state_dim_f: jnn.one_hot(s_f, state_dim_f), initial_state, [d.shape[-1] for d in env.params["D"])
self = eqx.tree_at(lambda x: x.params["D"], self, onehotted_states)
... rest of reset() logic
````

2) force users to pass in their own `initial_carry` to the rollout function if they reset their environment, so that line 182 of rollout is bypassed and the first use of `env` is in the `step_fn()` of rollout

Issue #273, opened 2025-09-15
Refactor `Env` (and `Agent`?) to be more functional and return a `state` like in JaxMARL and Jax+RL


conorheins:
Raised by @tverbele 

In the pymdp jax Env we currently keep the env state in the Env object and after rollout we get a batched version of this Env object back.
```python
observation_next, env_next = env.step(
            rng_key=keys[2:], actions=action_next
        ) 
```
Was looking a bit in how jax+RL is doing it, and for instance in gymnax / jaxMARL the flow is typically:

```python
key = jax.random.PRNGKey(0)
key, key_reset, key_act, key_step = jax.random.split(key, 4)

# Instantiate the environment & its settings.
env, env_params = gymnax.make("Pendulum-v1")

# Reset the environment.
obs, state = env.reset(key_reset, env_params)

# Sample a random action.
action = env.action_space(env_params).sample(key_act)

# Perform the step transition.
n_obs, n_state, reward, done, _ = env.step(key_step, state, action, env_params)
```

where the state is typically a separate object that gets passed (and that would end up in the info dict)...
Is there a rationale to keep everything packed in the Env equinox module, or should we consider moving to something more like these other jax environments use to be consistent?

Arun-Niranjan:
I think this refactor makes sense to do. It seems like it would have the following benefits:
1. more consistent with other JAX environments (as noted above)
2. testing our environment functions will in theory be easier as they will have fewer side-effects

tverbele:
From a pymdp / active inference perspective, I think it makes sense to have an Environment API like

```
obs, state = env.reset(key)
obs, state = env.step(key, state, action)
```
where `obs` would be an array of jnp arrays that you can feed into `Agent.infer_states` and `state` something environment specific. We could also provide wrapper implementations for some of the above environments to test active inference agents on these RL envs (e.g. converting the reward into an observation modality etc)

However, a few design decisions to make are:
- do we provide a key each call of `step` or let the env handle it (our current API does the former which lgtm)?
- do we consider a `done` signal and/or should the environment "auto-reset"?
- do we provide a way of specifying the goal in the environment, i.e. by specifying a C tensor or some reward signal after all?
- do we return `obs` or rather a more expressive `Timestep` dataclass object?
- do we provide a standard `render` method , i.e. `env.render(state)`
- do we assume env is batched (and reset/step are vmapped under the hood) like Agent, or rather have the user (or rollout) do a `jax.vmap(env.step)()` call?

Issue #259, opened 2025-09-10
Add option to construct `Env` using `A`, `B`, `C`, `D` and dependencies directly, rather than via `params` and `dependencies` dictionary


conorheins:
This would make the env mimic the `Agent` constructor syntax

Issue #219, opened 2025-08-19
Add unit tests for `rollout` function


conorheins:
We can re-use the `test_agent_jax.py` outputs which simulate active inference loop using a for-loop over time, as validation for the outputs of `rollout`

PR #334, opened 2025-12-19
Refactor environment class to make it more gymnax-like

conorheins:
This PR addresses issues #259 and #273 by refactoring the `Env` class to a functional, state‑passing API and introduces a sub-class `PymdpEnv` with `Agent`‑style construction -- based on `A`, `B`, `D`,  and `A_dependencies`, `B_dependencies`. 
The updates `rollout`, core envs, tests, and notebooks to match the refactor, and adds a new `test_env.py` file.

### Summary of changes
- Introduces an abstract `Env` API (registered as an [Abstract Base Class](https://docs.python.org/3/library/abc.html)). and a concrete `PymdpEnv` that takes `A`/`B`/`D` (or `Distribution`s) plus dependencies, with explicit `reset(key, state, env_params)` / `step(key, state, action, env_params)` and no env‑held mutable state. The `Env` / `PymdpEnv` classes now assumed single-instance key, action, state, and parameters, meaning that any batching / parallelization must be done manually by wrapping `jax.vmap` around the `.step()` and `.reset()` methods.
- Adds `pymdp.envs.env.make()` with `make_env_params` for returning `env_params`, and updates `generate_env_params` to return unbatched or batch‑broadcasted parameter pytrees. **Note**: `make(..., make_env_params=True)` returns `env_params` with the same shape as the input `A`/ `B` / `D`; use `env.generate_env_parameters(batch_size=...)`to create batched env_params from an initially single-instance or unbatched environment, explicitly.
- Updates `rollout` to vmap `env.reset/step` over batch, allow partial `initial_carry` overrides
- Refactors `PymdpEnv`‑based envs and demos (`GridWorld`, `TMaze`, `GraphEnv`, `GeneralizedTMaze`) to the new API
- Adds/updates tests for new env behavior and rollout consistency, plus a fix in `Agent` to avoid mutating shared `pA`/`pB` when broadcasting.

### Notable updates
- Tests: `test_env.py` (new `PymdpEnv` coverage), `test_rollout_function.py` (test for env_state history, addressing @Arun-Niranjan's request [here](https://github.com/infer-actively/pymdp/pull/280#discussion_r2385206660)), `test_grid_world_parity.py` (add test with batched `env_params`/state).
- Tutorials/demos: multiple notebooks updated to new API (graph world, grid world, t‑maze, generalized t‑maze, model construction, pybefit fitting demo).

### Side-effects on other open Issues
- This PR also adds a new unit test to `test_rollout_function.py` that should allow us to close out #219, by testing that the returned `state` from the info dict of `rollout` reflects the final state of the environment after multiple calls to `step`. 
- The ability to override the `initial_carry` with an initial state, and the passing of a `state` to the `reset` method, should also address/render-irrelevant Issue #279 .
- Putting all environmental parameters into a separate `env_params` pytree now also allows you to theoretically change the environmental parameters in between calls to `env.step()` and therefore dynamically manipulate the environmental parameters, addressing the request made in #165 

### Usage examples

In context of using a `PymdpEnv` in combination with rollout, we anticipate use-cases will fall into one of two main major types of use cases:

```python
# Case 1: static env parameters stored on the env
env = PymdpEnv(A=A, B=B, D=D, A_dependencies=..., B_dependencies=...)
agent = Agent(A, B, D=D, batch_size=BATCH, ...)

key = jr.PRNGKey(0)
last, info = rollout(agent, env, T, key)
# rollout internally vmaps env.reset/step across agent batch
```

```python
# Case 2: env params passed explicitly (batched to match agent.batch_size)
env = PymdpEnv(A=A, B=B, D=D, A_dependencies=..., B_dependencies=...)
agent = Agent(A, B, D=D, batch_size=BATCH, ...)

env_params = env.generate_env_params(batch_size=agent.batch_size)
last, info = rollout(agent, env, T, key, env_params=env_params)
```

When working with the environment's `step()` and `reset()` methods directly, you now have to vmap these methods manually over their input arguments, e.g.

```python
# batched manual step, over `keys`, `env_state`, `actions`,  and `env_params`
jax.vmap(env.step)(keys, env_state, actions, env_params=env_params)
```

### DIFF
--- a/pymdp/envs/env.py
+++ b/pymdp/envs/env.py
@@ -1,4 +1,5 @@
 from functools import partial
+from abc import ABC, abstractmethod
 
 import jax.numpy as jnp
 from jax import vmap, jit, random as jr, tree_util as jtu
@@ -68,11 +69,13 @@ def make(A, B, D, A_dependencies=None, B_dependencies=None, make_env_params=Fals
 
     return env, env_params
 
-class Env:
+class Env(ABC):
 
+    @abstractmethod
     def reset(self, key, state=None, env_params=None):
         raise NotImplementedError
 
+    @abstractmethod
     def step(self, key, state, action, env_params=None):
         raise NotImplementedError
 


Arun-Niranjan:
This is a large PR @conorheins, will take me a few days to review properly. Are there any parts you'd like particular attention on?

conorheins:
> This is a large PR @conorheins, will take me a few days to review properly. Are there any parts you'd like particular attention on?

Yeah good question -- I think the most useful place for feedback is just on the Env API side, which boils down to the changes made to `env.py` and `rollout.py`.

More specifically, the main API changes I would like to get feedback on are the following:
1. **Putting `env_params` into a separate dictionary**. Now with the `env_params` option, we can put the environmental parameters (A/B/D) into an `env_params` dict which is more functional/aligned with other JAX RL frameworks. In this case the Env class becomes a thin wrapper over the `.step()` and `.reset()` methods which take `key`, `action`, `state`, `env_params`.
2. **How we handle batching with `env_params`  and `rollout`**. With the new changes, whenever you pass `env_params` to the `rollout` method, you have to match the batch dimension of `env_params` to the batch dimension of `Agent`. For a while I was going back and forth with @tverbele about having the option of having _unbatched_ `env_params`  (allowing the case of multiple agents, but one environment), but we decided against it because of the extra argument complexity this would add to the `rollout()` function, requiring another flag like `batched_env_params=[True/False]`. So right now, the natural way to do this (unbatched env, but batched agents), is to load all the parameters into the `Env` class itself (e.g., as `self.A`, `self.B`, `self.D` in case of a `PymdpEnv` class) and leave `env_params=None` in the arguments to `rollout`. The downside of this, is that we can't do functional/online changes to `env_params` (as requested in Issues like #165) in the case of using `rollout()` with batched agents but unbatched envs, like we can do in the batched agents, batched envs case.
3. **The general structure of `PymdpEnv`**

riddhipits (changes requested):
The unit tests ran successfully. To thoroughly test the refactored environment class, I created a simple environment here with different experiments to test different features of the refactorisation: [testing_refactored_env.ipynb](https://github.com/user-attachments/files/24374218/testing_refactored_env.ipynb). 

The general structure of `PymdpEnv` and introduction of `env_params` and `env_states` to be more like other JAX RL environments is good and clear. The way batching is handled for `env_params` and `rollout` makes sense. Yes, having to set `env_params` as `None` when the environment is shared among agents is a limitation for dynamic shared environments that require updating of `env_params` in `env.step`, but I think it's okay to not address this limitation now. If `env_params` need to be updated mid-rollout, users can just treat them as blocks of trials by creating new environment instances and passing the `env_state` and `observation` from the previous rollout's `last` carry as `initial_carry` to the new rollout, and optionally allowing new agents to be initialised. I've provided an example of what I mean in `testing_refactored_env.ipynb` in the section titled "## testing if we can create a new `env` instance for dynamic shared environments" where the environment's B is updated to allow wrapped location transitions and the agents' B and C are updated accordingly.

A problem I spotted is in the generation of `env_params` in make(). When `make_env_params=True` and `batch_size` is provided, the function ignores the batch dimension entirely. The `make()` function directly converts A, B, and D to arrays without using the `generate_env_params()` method, which properly handles batching. This results in unbatched parameters. We can simply replace lines 59-64 in `make()` as follows: 
`if make_env_params:
           batch_size = kwargs.get('batch_size', None)
           env_params = env.generate_env_params(batch_size=batch_size)
else:
           env_params = None`

Issue #336, opened 2025-12-30
`get_likelihood_single_modality` computes expected probability instead of expected log probability for soft observations


conorheins:
When `distr_obs=True`, `get_likelihood_single_modality` currently computes the **expected probability** of observations under the likelihood matrix, rather than the **expected log-likelihood** required for variational message passing / fixed-point updates of hidden states.

Specifically, the current implementation:

```python
expanded_obs = jnp.expand_dims(o_m, tuple(range(1, A_m.ndim)))
likelihood = (expanded_obs * A_m).sum(axis=0)
```

computes:

$$
\mathbb E_{q(o)}[p(o \mid s)] = \sum_o q(o),A[o,s]
$$

When this quantity is later logged and used as a message to hidden states, this yields:

$$
\log \mathbb E_{q(o)}[p(o \mid s)]
$$

However, the correct variational backward message from the likelihood factor to hidden states is:

$$
\mathbb E_{q(o)}[\log p(o \mid s)] = \sum_o q(o),\log A[o,s]
$$

These are not equivalent in general (by Jensen’s inequality), and this discrepancy leads to incorrect state updates when observations are represented as categorical distributions rather than one-hot indices.

---

### Why this matters

For discrete active inference models using variational message passing:

* One-hot observations correctly contribute $log A[o, s]$
* Soft (distributional) observations should contribute $E_{q(o)}[log A[o, s]]$
* The current implementation instead treats soft observations as a mixture over probabilities, which is inconsistent with the variational objective

This affects:

* fixed-point iteration for posterior state inference
* any setting using `distr_obs=True` with non-one-hot categorical observations

---

### Expected behavior

For `distr_obs=True`, the likelihood message to hidden states should be computed as:

```python
(expanded_obs * jnp.log(A_m + eps)).sum(axis=0)
```
---

### Possible resolution

Options include:

1. Returning the expected log-likelihood directly when `distr_obs=True`
2. Renaming the current function to clarify that it computes $E[p(o|s)]$, and introducing a separate VMP-correct log-likelihood message function
3. Auditing downstream call sites to determine whether they expect probabilities or log-messages

PR #337, opened 2026-01-05
Fix distributed obs likelihood

Arun-Niranjan:
Addresses https://github.com/infer-actively/pymdp/issues/336 - see this issue for a very detailed explanation of why the old maths was wrong and why this is correct.

We don't have any notebooks/applications which are using distributed observations that are not one-hot (yet), so there are no other downstream changes required.

### DIFF
--- a/pymdp/maths.py
+++ b/pymdp/maths.py
@@ -105,7 +105,7 @@ def compute_log_likelihood_single_modality(o_m, A_m, distr_obs=True):
     """Compute observation log-likelihood for a single modality"""
     if distr_obs:
         expanded_obs = jnp.expand_dims(o_m, tuple(range(1, A_m.ndim)))
-        log_likelihood = jnp.sum(expanded_obs * jnp.log(A_m + MINVAL), axis=0)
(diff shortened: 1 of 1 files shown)


pymdp/maths.py
```diff
@@ -101,19 +101,14 @@ def factor_dot_flex(M, xs, dims: List[Tuple[int]], keep_dims: Optional[Tuple[int
...
 
 
-def get_likelihood_single_modality(o_m, A_m, distr_obs=True):
-    """Return observation likelihood for a single observation modality m"""
+def compute_log_likelihood_single_modality(o_m, A_m, distr_obs=True):
+    """Compute observation log-likelihood for a single modality"""
     if distr_obs:
         expanded_obs = jnp.expand_dims(o_m, tuple(range(1, A_m.ndim)))
-        likelihood = (expanded_obs * A_m).sum(axis=0)
+        log_likelihood = jnp.sum(expanded_obs * jnp.log(A_m + MINVAL), axis=0)
```
conorheins:
```suggestion
        log_likelihood = xlogy(expanded_obs, jnp.clip(A_m, min=MINVAL)).sum(axis=0)
```

For numerical stability

Issue #335, opened 2025-12-21
CI OOM flakiness in `test_infer_states_optimized.py` due to unseeded random spec generation + exponential end2end padding


conorheins:
**Summary:** 
Our CI unit tests are intermittently failing with a JAX `RESOURCE_EXHAUSTED` OOM (e.g. allocating ~20 GiB) in `test/test_infer_states_optimized.py`, specifically inside `apply_A_end2end_padding_batched()`. The test generates agent specs using unseeded `np.random` via `generate_agent_specs_from_parameter_sets()`, so the exact spec set varies run-to-run; occasionally we sample a spec with a long `A_dependencies` list (large max dependency rank) and moderate `max_state_dim`, which makes the end2end padding allocate an enormous dense tensor (scales like `Smax^K`). This leads to CI flakiness: rerunning the same workflow can “fix” it by sampling a luckier spec.

**Proposed fix:** 
Make the tests deterministic (seed the spec generator) and add a guard/skip for specs whose estimated padded allocation exceeds a threshold (or cap dependency length for the test spec generator). Optionally update CI xdist settings (e.g. `--dist=loadscope`) to reduce multi-worker randomness effects, and consider fixing the spec generator logic where dependency-length sampling uses a factor index instead of that factor’s dimensionality.

PR #339, opened 2026-01-07
Fix oom kill unit test

Arun-Niranjan:
Addresses #335 

Changes include:
- adding the option to specify a random seed when generating an agent spec
- adding an option to specify the maximum A dependency depth limit
- using these options in the unit test that was randomly OOM-ing.
- using JAX instead of numpy for the random number generation

~~Default behaviour remains unchanged~~
Note: in keeping with JAX norms, the default behaviour is now changed. Instead of non-deterministic by default, we assume a key of 0 if none is provided. Because these methods are only used in the tests (which passed in a seed before) this is not a breaking change to anything else in pymdp

### DIFF
--- a/pymdp/utils.py
+++ b/pymdp/utils.py
@@ -347,7 +347,7 @@ def generate_agent_spec(
         A_dep_len_limits,
         dim_sampling_type,
         A_dep_len_prior='uniform',
-        rng=None,
+        key=None,
     ):
 
     '''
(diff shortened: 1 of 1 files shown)


pymdp/utils.py
```diff
@@ -365,13 +368,18 @@ def generate_agent_spec(num_factors,
     A_dep_len_prior:   'uniform'| 'exponential' - whether to sample lengths of A
                        dependency lists uniformly or to favor shorter ones, in cases
                        where no hidden states have already been assigned to a list
+    rng:               numpy random Generator or RandomState, for reproducible random
+                       number generation. If None, uses default_rng()
     '''
+
+    if rng is None:
```
conorheins: Can you refactor this to use jax PRNGKeys here instead of numpy random keys? I think that anytime we have an opportunity to change something related to random seeding, we should refactor if possible to using jax.random

test/test_infer_states_optimized.py
```diff
@@ -73,7 +77,9 @@ def setUpClass(cls):
         spec_data = generate_agent_specs_from_parameter_sets(
             parameter_sets,
             num_agents_per_set=1,
-            output_file=None  # Don't save to file
+            max_A_dependency_list_size=3,
```
conorheins, line 80: waiting on confirmation from @nikolamilovic-ft to see whether reducing this to 3 is valid/sufficient for validating the behavior of the optimized inference methods

Arun-Niranjan: There's a case to be made that if this test *needs* to be using a ton of memory, it should be a benchmark that runs on a slower cadence than the standard unit tests

Arun-Niranjan: I'm going to leave this as it is for now to ensure we don't have flaky CI - happy to create a ticket for adding memory and speed benchmark tests that can be run in a script locally, or perhaps we fork out for more powerful github action runners that run once a week

conorheins: Sounds good, got confirmation that the max_A_dependency_list_size=3 is sufficient for establishing that the methods work as intended anyway

PR #341, opened 2026-01-16
Jitted sophisticated inference planning algorithm

riddhipits:
This is a complete rewrite of the sophisticated inference planning algorithm with JAX integration, algorithm optimisations, a visualisation suite, and a validation suite.

The key changes include: 
- Full rewrite using JAX with JIT compilation for performance gains
- Replaced Python lists with pre-allocated JAX arrays using `equinox.Module`
- Batched processing with `vmap` and `lax.scan` throughout
- Introduced new features summarised in the documentation under `optimized_tree_search()` in `si.py`
- Plotting of the planning tree in `visualize.py` 
- Validation suite in `si_tmaze_SIvalidation.ipynb` using the T-Maze task as in the Sophisticated Inference paper to compare between vanilla active inference and sophisticated inference planning algorithms.

Question: Currently, in the SI planning algorithm, we set the default gamma value to be 1.0 to ensure less aggressive pruning (so users don't unintentionally prune aggressively and get confused about results) but the rest of the pymdp repo uses the default value of 16.0 for the Agent and vanilla planning algorithm, so should we make the SI planning algorithm's gamma value to also be 16.0 by default for consistency?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,6 @@ dependencies = [
     'seaborn>=0.11.1',
     'mctx>=0.0.5',
     'networkx>=3.3',
-    "pygraphviz>=1.14",
 ]
 
 [dependency-groups]


pyproject.toml
```diff
@@ -32,6 +32,7 @@ dependencies = [
     'seaborn>=0.11.1',
     'mctx>=0.0.5',
     'networkx>=3.3',
+    "pygraphviz>=1.14",
```
Arun-Niranjan: I don't think you need this in the main dependencies as we have it in the `nb` (notebook) dependency group (you can run `uv sync --all-groups`) and that will handle it for you

Arun-Niranjan: long term I'd like to try and decouple all of the rendering, animation and visualisation code from the main pymdp package - to improve package portability and make it easier to deploy in production environments

riddhipits:
> > Question: Currently, in the SI planning algorithm, we set the default gamma value to be 1.0 to ensure less aggressive pruning (so users don't unintentionally prune aggressively and get confused about results) but the rest of the pymdp repo uses the default value of 16.0 for the Agent and vanilla planning algorithm, so should we make the SI planning algorithm's gamma value to also be 16.0 by default for consistency?
> 
> If gamma is used differently in sophisticated inference relative to other methods, perhaps it needs a different name?

It's not being used differently though

conorheins:
Not finished with review yet - but just added some unit tests for sophisticated inference in `test/` that test for the following behaviors (qualitative):
1. `test_si_accepts_costly_informative_cue`: SI planning chooses the costly cue action (index 3) because it’s information‑seeking despite negative immediate utility, while the vanilla policy search with utility + state info‑gain avoids that cue and prefers staying.
2. `test_si_ignores_irrelevant_distractor_cue`: In a dual‑cue setup, SI prefers the informative cue A over the irrelevant cue B, whereas the vanilla search treats both cues as equally attractive (action marginals roughly equal).

riddhipits:
> Not finished with review yet - but just added some unit tests for sophisticated inference in `test/` that test for the following behaviors (qualitative):
> 
> 1. `test_si_accepts_costly_informative_cue`: SI planning chooses the costly cue action (index 3) because it’s information‑seeking despite negative immediate utility, while the vanilla policy search with utility + state info‑gain avoids that cue and prefers staying.
> 2. `test_si_ignores_irrelevant_distractor_cue`: In a dual‑cue setup, SI prefers the informative cue A over the irrelevant cue B, whereas the vanilla search treats both cues as equally attractive (action marginals roughly equal).

Hey Conor, thanks for these unit tests. Quick question about the second one, is the dual-cue setup in the case of using a generalised t-maze or the current one used for SI validation (si_tmaze)? Because the latter only accommodates one cue I believe.

conorheins:
> > Not finished with review yet - but just added some unit tests for sophisticated inference in `test/` that test for the following behaviors (qualitative):
> > 
> > 1. `test_si_accepts_costly_informative_cue`: SI planning chooses the costly cue action (index 3) because it’s information‑seeking despite negative immediate utility, while the vanilla policy search with utility + state info‑gain avoids that cue and prefers staying.
> > 2. `test_si_ignores_irrelevant_distractor_cue`: In a dual‑cue setup, SI prefers the informative cue A over the irrelevant cue B, whereas the vanilla search treats both cues as equally attractive (action marginals roughly equal).
> 
> Hey Conor, thanks for these unit tests. Quick question about the second one, is the dual-cue setup in the case of using a generalised t-maze or the current one used for SI validation (si_tmaze)? Because the latter only accommodates one cue I believe.

The second "dual-cue" one is inspired by the generalized tmaze, yes -- where you have one information-rich but irrelevant cue (it reveals location of a reward-neutral hidden state), and one information-rich _and relevant_ cue (it reveals the location of a rewarding state). The SI agent should be picky foraging information from the information rich _and_ relevant cue (unlike the vanilla AIF agent), due to the uniquely-SI counterfactual belief-updating that occurs when exploring (action, observation) branches conditioned upon visiting that cue. The first unit test is more inspired by your SI validation jupyter notebook (in case of uncommenting that code that would add cost to the cue-observation), because it shows how an SI agent, rather than a vanilla AIF agent, will choose an informative cue despite the cost

Arun-Niranjan:
> > > Not finished with review yet - but just added some unit tests for sophisticated inference in `test/` that test for the following behaviors (qualitative):
> > > 
> > > 1. `test_si_accepts_costly_informative_cue`: SI planning chooses the costly cue action (index 3) because it’s information‑seeking despite negative immediate utility, while the vanilla policy search with utility + state info‑gain avoids that cue and prefers staying.
> > > 2. `test_si_ignores_irrelevant_distractor_cue`: In a dual‑cue setup, SI prefers the informative cue A over the irrelevant cue B, whereas the vanilla search treats both cues as equally attractive (action marginals roughly equal).
> > 
> > 
> > Hey Conor, thanks for these unit tests. Quick question about the second one, is the dual-cue setup in the case of using a generalised t-maze or the current one used for SI validation (si_tmaze)? Because the latter only accommodates one cue I believe.
> 
> The second "dual-cue" one is inspired by the generalized tmaze, yes -- where you have one information-rich but irrelevant cue (it reveals location of a reward-neutral hidden state), and one information-rich _and relevant_ cue (it reveals the location of a rewarding state). The SI agent should be picky foraging information from the information rich _and_ relevant cue (unlike the vanilla AIF agent), due to the uniquely-SI counterfactual belief-updating that occurs when exploring (action, observation) branches conditioned upon visiting that cue. The first unit test is more inspired by your SI validation jupyter notebook (in case of uncommenting that code that would add cost to the cue-observation), because it shows how an SI agent, rather than a vanilla AIF agent, will choose an informative cue despite the cost

I know we're in the weeds of Jax, python and api designs - but I just want to say this is unbelievably cool

conorheins (approved):
This PR introduces a JITTed implementation of sophisticated inference which is an excellent, important and long-needed addition to `pymdp`. 

I did some cleanup / new additions in the course of reviewing the PR, which include the following:

- refactored the `TMaze` environment into a single base class `BaseTMaze` (in the current `envs/tmaze.py` file) which then has specific wrappers for the default `TMaze` and the new `SimplifiedTMaze`  (formerly in si_tmaze.py). `SimplifiedTMaze` is the reduced version of the T-Maze environment which has fully-reachable (all-to-all) transition matrix connectivity between states, and fewer states (no "middle" state between the two arms). It also collapses cue and location observations into a single modality. 
- Use the `SimplifiedTmaze` environment in the `si_tmaze_SIvalidation.ipynb` notebook for comparison with the experiments from the original [paper](https://pubmed.ncbi.nlm.nih.gov/33626312/) which introduced sophisticated inference. 
- added an example of using sophisticated inference to show the same principle using the original `TMaze` environment (not the simplified one). I had to increase `observation_prune_threshold` from 0.0 to 1e-4 to allow the search / planning tree to be small enough to allow the intended behavior to appear.
- added unit tests for sophisticated inference (qualitative behavior checks on SI behavior in two cases: one in a simple state space with distracting cues, and another in a case where there's a single cue but has some cost associated with it, that only the SI agent can 'see through')
-  Updated the `si_generalized_tmaze.ipynb` and `si_graph_world.ipynb` demo notebooks to use the new observation/action description features introduced in the `visualize.py` file, so that the planning trees are annotated with observation and action labels.

Issue #297, opened 2025-10-29
Add variational free energy computation


conorheins:
Add a method that computes the variational free energy for a given pair of (generative model, variational posterior).

A few different options that I can imagine

- computing the 'single timestep' variational free energy (effectively computed currently inside the vanilla fixed-point iteration algorithm) using only the current posterior over hidden states, past action, B matrix, A matrix and priors over A and B. There is a version of this already in the `numpy` backend ([pymdp.legacy.maths::calc_free_energy()](https://github.com/infer-actively/pymdp/blob/fcd483fc5dd8c896af4ca1cebdd3e8c9e0c80800/pymdp/legacy/maths.py#L396C5-L396C21)) but it needs to be generalized and KL divergences for parameter posteriors/priors need to be added
- computing the full variational free energy given a sequence of observations, actions, hidden state posteriors (indexed from t=0 to t=current timestep) and parameter posteriors/priors.

In the case of the latter (full free energy), I guess we should have some check that makes sure that if the inference algorithm is "ovf" or "fpi", that smoothing should be performed before computing the full variational free energy.

Issue #189, opened 2025-07-03
Have control functions return EFE, not negative EFE?


conorheins:
More of a discussion point, but should we have the functions in the `control.py` library return the non-inverted, "original" expected free energy rather than the negative expected free energy?

tverbele:
I think `G` should be EFE, or keep returning negative EFE for backwards/spm compatibility reasons but call it `neg_G` instead?

PR #357, opened 2026-02-28
Add preprocessing and SVD image discretisation for RGM MNIST classifier notebook

Arun-Niranjan:
This change creates a notebook for demonstrating how to construct an MNIST RGM classifier as per the ["From pixels to planning" paper](https://www.frontiersin.org/journals/network-physiology/articles/10.3389/fnetp.2025.1521963/full).

The notebook up to this point does everything required before the construction of pymdp agents, including:
1. loading MNIST images
2. preprocessing
3. singular value decomposition in order to discretise images

I've tried to keep as much "business" logic in functions outside the notebook, which are then imported and results plotted to make the code easier to review.

This is my best reproduction of what is described in the paper, although I did find a few contradictions between the main text and the footnotes/figure captions. (number of components, whether groups overlap).

I do notice that my reconstructions don't look as nice as figure 4. the discretisation down to 7 levels per singular variate seems pretty harsh.

future PRs will add:
- fast structure learning (DONE):
  - construction of pymdp agents that map the discretised images to hidden states [DONE]
  - stacking of the agents in such a way that they eventually map to digit classes [DONE]
- active parameter learning:
  - using the free energy to decide on gating parameter updates


uv.lock
conorheins, line 1: In general we have not been committing `uv.lock` files because it leads to a lot of commit churn, due to which dep groups people will be using and which accelerator backend (CPU vs GPU), which will lead to this changing a lot. Can you remove this from the commit?

pyproject.toml
```diff
@@ -50,8 +54,8 @@ test = [
 
 [project.optional-dependencies]
 gpu = [
-    'jax[cuda12]>=0.3.4',
-    'jaxlib[cuda12]>=0.3.4',
+    "jax[cuda13]>=0.6.2",
```
conorheins, line 57: any reason you upgraded cuda version here?

conorheins (commented):
Nice! This looks like a promising start to the RGM MNIST classification demo.  The preprocessing workflow is very cleanly implemented and clear.

First, some low-level comments:

### Low level comments

1) can you please rename to _renormalizing_ generative models where it appears, instead of _renormalised_? 

2) Have you tried comparing performance (at this point, I just mean the preprocessing SVD reconstruction stage), against the demos from the public SPM release: https://github.com/spm/spm/blob/main/toolbox/DEM/DEM_MNIST.m and https://github.com/spm/spm/blob/main/toolbox/DEM/DEM_MNIST_RGM.m? 
That might be one way to  determine the source of the discrepancies you mentioned (like # of singular vectors needed for reconstruction). It could be that the underlying SVD code is slightly different in some way.

Now, some high-level thoughts related to threading this into the 1.0.0 release:

### High level planning considerations
I think we should only include this in the official `examples/` folder for the 1.0.0 release if it can do the RGM MNIST classification demo from end to end, achieving performance at least equivalent to what is reported in the original paper. This is in line with our decision during the notebook_cleanup push from the summer, which was only to include fully end-to-end demos. This is for instance why we removed the old pybefit demos and had to finish implementing a few others (e.g., the `learning_gridworld.ipynb` notebook, which was a WIP at the time).

Given that, I would recommend that we only merge this and the follow-up PRs in, if you are confident you can get the full demo working by the time 1.0.0 is ready to be released. Otherwise, if we want to keep this as an ongoing / experimental development project, I suggest we push it to a branch that is rebased on the v1.0.0_alpha branch (soon to be: master branch) but not actually merged to master until it's working. 

In order to help you answer that^, in terms of time-to-1.0.0 release : the documentation PR in #355 and a few other tickets (#189, #196, #277, and #297) are the only ones needing addressing before the 1.0.0 release, and I am confident I can address them within 2-3 weeks, at which point releasing 1.0.0 will be unblocked.

Given that, do you think you would be able to finish off the MNIST RGM demo in parallel within the next 2-3 weeks? If so, then we should go ahead and get this into 1.0.0 because it's a nice demonstration / example to have, but otherwise I say we go with the 'add to experimental/dev branch' option which doesn't require a fully functional demo within the next 2-3 weeks.

pyproject.toml
Arun-Niranjan: I was getting errors along the lines of "cuda12 not found" - but will remove and do an upgrade as a separate PR.

uv.lock
Arun-Niranjan: yes will do! my understanding is that uv.lock files should be committed for applications, but not libraries

examples/renormalised_generative_models/discretise.py
```diff
@@ -0,0 +1,170 @@
...
+    """Radial Gaussian weighting mask for a flattened patch (sigma = group_size)."""
+    g = config.group_size
+    x = jnp.arange(g) - (g - 1) / 2
+    xx, yy = jnp.meshgrid(x, x)
+    sigma = float(g)
+    w2d = jnp.exp(-(xx**2 + yy**2) / (2 * sigma**2))
+    return jnp.tile(w2d, (config.n_channels, 1, 1)).flatten()
+
+
+def partition_into_patches(images: jnp.ndarray, config: DiscretiseConfig) -> jnp.ndarray:
```
conorheins, line 48: One thing that might be the source of the discrepancy you're noticing is this patching logic (lines 40-46) -- if I understand correctly, this partitions the images into nonoverlapping 4x4 patches (since `config.group_size` defaults to 4), whereas in SPM the patches are going to be overlapping -- see here for instance in `spm_tile.m` which is the main patching function used in the RGM demos, as far as I can tell/remember: https://github.com/spm/spm/blob/066c22d4f1d2a2a88066bd91db0acf250edc1d5c/toolbox/DEM/spm_tile.m#L48

examples/renormalised_generative_models/preprocess.py
```diff
@@ -0,0 +1,121 @@
...
+        return cdf_normalized[indices].reshape(img.shape)
+
+    return vmap(equalize_single)(images)
+
+def convert_to_rgb(images: jnp.ndarray) -> jnp.ndarray:
+    return jnp.stack([images, images, images], axis=1)
+
+
+@jax.jit(static_argnames=['target_size', 'sigma'])
+def preprocess(images: jnp.ndarray, target_size: int = 32, sigma: float = 0.3) -> jnp.ndarray:
```
conorheins: the preprocessing pipeline here also seems to deviate from what is done in [`DEM_MNIST_RGM.m`](https://github.com/spm/spm/tree/main/toolbox/DEM/DEM_MNIST_RGM.m), where the upsampling is done first, then smoothing, then rank-remapping of the histogram, see lines [54](https://github.com/spm/spm/blob/066c22d4f1d2a2a88066bd91db0acf250edc1d5c/toolbox/DEM/DEM_MNIST_RGM.m#L54) - [59](https://github.com/spm/spm/blob/066c22d4f1d2a2a88066bd91db0acf250edc1d5c/toolbox/DEM/DEM_MNIST_RGM.m#L59)

examples/renormalised_generative_models/discretise.py
```diff
@@ -0,0 +1,170 @@
...
+    Args:
+        images: (N, C, H, W) preprocessed structure images
+        config: discretisation parameters
+
+    Returns:
+        SVDBasis with learned projection and quantisation parameters
+    """
+    g = config.group_size
+    n = config.image_size // g
+    k = config.n_components
```
conorheins: Another possible source of discrepancy: it appears in [`spm_rgb2O.m`](https://github.com/spm/spm/blob/066c22d4f1d2a2a88066bd91db0acf250edc1d5c/toolbox/DEM/spm_rgb2O.m#L130), the SVD step first thresholds the normalized singular values then caps them by the `mm` variable. The thresholding is happening internally in the function [`spm_svd.m`](https://github.com/spm/spm/blob/main/spm_svd.m), and the threshold argument here is `1/su` (second arg to `spm_svd`). This means only singular values/vectors are kept, whose normalized singular value exceeds `1/su`. This has the effect of keeping fewer singular vectors in "simple" (e.g., background) patches and more in complex patches.

```matlab
    % singular value decomposition for this group
    %----------------------------------------------------------------------
    Y        = times(double(I(:,G{g})),W{g});
    [u,s,v]  = spm_svd(Y,1/su);
    Nm       = min(length(s),mm);
    if Nm
        RGB.V{g} = v(:,1:Nm);
        u        = u(:,1:Nm)*s(1:Nm,1:Nm);
    end
```

examples/renormalised_generative_models/discretise.py
```diff
@@ -0,0 +1,170 @@
...
+"""
+
+from typing import NamedTuple
+
+from jax import numpy as jnp, vmap
+
+
+class DiscretiseConfig(NamedTuple):
+    group_size: int = 4
+    n_components: int = 16
```
conorheins: The discrepancy you noted between the paper mentioning 32 vs 16 singular values is real, but it seems in practice the maximum number of singular values kept per patch is going to be 16, see [line 86 in the RGM demo script.](https://github.com/spm/spm/blob/066c22d4f1d2a2a88066bd91db0acf250edc1d5c/toolbox/DEM/DEM_MNIST_RGM.m#L86)

examples/renormalised_generative_models/discretise.py
```diff
@@ -0,0 +1,210 @@
...
+    centred_2d = centred.transpose(1, 2, 0, 3).reshape(n * n, N, f)
+
+    def svd_one(data):
+        _U, S_full, Vh = jnp.linalg.svd(data, full_matrices=False)
+        V = Vh.T[:, :k]
+        S = S_full[:k]
+
+        # Adaptive mode selection: retain modes where S[i]/S[0] > 1/sv_threshold
+        threshold = 1.0 / config.sv_threshold
+        normalized = S / jnp.maximum(S[0], 1e-12)
```
conorheins: I think this threshold is close in spirit to SPM but not equivalent.
Here it seems you're keeping eigenvectors in terms of the their ratio to the top singular values, like `s_i / s_1`

In `spm_svd`, the criterion is normalized by total spectrum (mean energy), e.g.:
`lambda_i * len(lambda) / sum(lambda) > 1/su`   (with lambda_i = s_i^2 in the m>n conditional case), then capped by `mm`. So this line can under-retain modes vs SPM for the same `su`, especially when the first singular value is dominant.

examples/renormalised_generative_models/discretise.py
```diff
@@ -0,0 +1,210 @@
...
+    # (N, n, n, f) -> (n*n, N, f) for vmapped SVD
+    centred_2d = centred.transpose(1, 2, 0, 3).reshape(n * n, N, f)
+
+    def svd_one(data):
+        _U, S_full, Vh = jnp.linalg.svd(data, full_matrices=False)
+        V = Vh.T[:, :k]
+        S = S_full[:k]
+
+        # Adaptive mode selection: retain modes where S[i]/S[0] > 1/sv_threshold
+        threshold = 1.0 / config.sv_threshold
```
conorheins, line 107:
```suggestion
        threshold = 1.0 / config.sv_threshold
        eigvals = jnp.square(S)
        L = eigvals.shape[0]
        normalized = eigvals * (L / jnp.maximum(eigvals.sum(), 1e-12))
        mode_mask = (normalized > threshold).astype(jnp.float32)

        # Keep at least one mode for near-degenerate patches.
        mode_mask = mode_mask.at[0].set(jnp.where(mode_mask.sum() > 0, mode_mask[0], 1.0))
        n_modes = mode_mask.sum().astype(jnp.int32)
```

(sorry, i dont know how to do multi-line code suggestions from the GH UI, but this is suggested to replace lines 103-107)

examples/renormalised_generative_models/discretise.py
Arun-Niranjan: yes I definitely found the paper description + figure 4 confusing, as it's not obvious whether TxT refers to the dimensions of the patch or the number of patches.

I've added overlapping tiling as well now and will do a comparison between the two, but will later drop the non-overlapping patches

conorheins: Yeah agreed it is confusing -- I remember when @tverbele @alec-tschantz and I were implementing RGMs (this was for the dove reconstruction video though, not the MNIST classification one), we convinced ourselves that it's definitely overlapping patches, for at least that example

Arun-Niranjan: tbh it does make sense relative to the later parts of the discussion talking about receptive fields

PR #393, opened 2026-04-22
fix: replace mutable default `shape=[2,2]` with `None` guard in grid_worlds.py

### LINKED ISSUE #392: Bug: mutable default argument `shape=[2, 2]` in GridWorldEnv and DGridWorldEnv constructors
### Summary
In `pymdp/legacy/envs/grid_worlds.py`, both `GridWorldEnv.__init__` and `DGridWorldEnv.__init__` use a mutable list as a default argument:

```python
def __init__(self, shape=[2, 2], init_state=None):
```

Using a mutable default argument is a well-known Python anti-pattern (B006 in flake8-bugbear). Because Python evaluates default argument values once at function definition time, any in-place mutation of the `shape` list in one call would silently affect all subsequent calls that rely on the default. While the current code does not mutate `shape` in-place, it is a latent bug that can be triggered by future changes and is flagged by static analysis tools.

### Fix
Replace the mutable default with `None` and assign the default inside the function body:

```python
def __init__(self, shape=None, init_state=None):
    if shape is None:
        shape = [2, 2]
```

This makes the intent explicit and eliminates the risk of unintended mutation across instances.

Signed-off-by: lingxiu58 <86288566+lingxiu58@users.noreply.github.com>

lingxiu58:
## Summary

Fixes #392

Both `GridWorldEnv` and `DGridWorldEnv` in `pymdp/legacy/envs/grid_worlds.py` used a mutable list `[2, 2]` as a default argument:

```python
def __init__(self, shape=[2, 2], init_state=None):
```

Using a mutable default is a Python anti-pattern (PEP B006). Python evaluates default argument values **once** at function definition time, so any in-place mutation of the shared list object would silently affect all future calls that rely on the default. Although the current code does not mutate `shape` in-place today, this is a latent bug and flagged by static analysis.

## Fix

Replace the mutable default with `None` and assign the list inside the function body:

```python
def __init__(self, shape=None, init_state=None):
    if shape is None:
        shape = [2, 2]
```

Applied to both `GridWorldEnv.__init__` and `DGridWorldEnv.__init__`.

Signed-off-by: lingxiu58 <86288566+lingxiu58@users.noreply.github.com>

### DIFF
--- a/pymdp/legacy/envs/grid_worlds.py
+++ b/pymdp/legacy/envs/grid_worlds.py
@@ -26,7 +26,7 @@ class GridWorldEnv(Env):
 
     CONTROL_NAMES = ["UP", "RIGHT", "DOWN", "LEFT", "STAY"]
 
-    def __init__(self, shape=[2, 2], init_state=None):
+    def __init__(self, shape=None, init_state=None):
         """
         Initialization function for 2-D grid world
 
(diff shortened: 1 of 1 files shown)


conorheins (approved):
Thanks for this @lingxiu58 ! looks good to me

PR #395, opened 2026-04-22
fix(maths): clip y in stable_cross_entropy to prevent NaN/-inf

### LINKED ISSUE #394: Bug: `stable_cross_entropy` does not clip `y` before `log`, risking NaN/-inf
### Description

In `pymdp/maths.py`, the `stable_cross_entropy` function computes:

```python
def stable_cross_entropy(x: ArrayLike, y: ArrayLike) -> ArrayLike:
    return - xlogy(x, y).sum()
```

This calls `xlogy(x, y)` which computes `x * log(y)`. If `y` contains values at or near zero, `log(y)` will produce `-inf` or `NaN`, leading to numerical instability.

By contrast, the closely related `stable_xlogx` function correctly clips its input:

```python
def stable_xlogx(x: ArrayLike) -> ArrayLike:
    return xlogy(x, jnp.clip(x, MINVAL))
```

`stable_cross_entropy` is missing this protective clipping of `y`.

### Reproduction

```python
import jax.numpy as jnp
from pymdp.maths import stable_cross_entropy

x = jnp.array([0.5, 0.5])
y = jnp.array([0.0, 1.0])  # y has a zero
print(stable_cross_entropy(x, y))  # -inf or NaN
```

### Fix

Clip `y` before passing to `xlogy`, consistent with how `MINVAL` is used elsewhere in this file:

```python
def stable_cross_entropy(x: ArrayLike, y: ArrayLike) -> ArrayLike:
    return - xlogy(x, jnp.clip(y, MINVAL)).sum()
```

### Impact

`stable_cross_entropy` is called inside `calc_vfe` to compute the prior cross-entropy term, so this bug can cause VFE computations to return `-inf` or `NaN` whenever the prior distribution `p_f` contains zero-probability states.

lingxiu58:
## Summary

Fixes #394

`stable_cross_entropy(x, y)` passes `y` directly to `xlogy(x, y)`, which computes `x * log(y)`. When `y` contains zero-probability values this produces `-inf` or `NaN`, silently corrupting VFE calculations.

Every other numerically sensitive function in `maths.py` already clips to `MINVAL` before logging (`stable_xlogx`, `log_stable`, `dirichlet_expected_value`, …). This one-line fix brings `stable_cross_entropy` in line with that convention.

## Change

```python
# Before
return - xlogy(x, y).sum()

# After
return - xlogy(x, jnp.clip(y, MINVAL)).sum()
```

## Testing

The fix is a one-line change identical in structure to the existing `stable_xlogx` implementation, which already has test coverage. A quick sanity check:

```python
x = jnp.array([0.5, 0.5])
y = jnp.array([0.0, 1.0])
print(stable_cross_entropy(x, y))  # Before: -inf | After: finite value
```

---
Signed-off-by: lingxiu58 <86288566+lingxiu58@users.noreply.github.com>

### DIFF
--- a/pymdp/maths.py
+++ b/pymdp/maths.py
@@ -59,7 +59,7 @@ def stable_cross_entropy(x: ArrayLike, y: ArrayLike) -> ArrayLike:
     ArrayLike
         Cross-entropy value.
     """
-    return - xlogy(x, y).sum()
+    return - xlogy(x, jnp.clip(y, MINVAL)).sum()
 
 def log_stable(x: ArrayLike) -> ArrayLike:
     """Compute stable logarithm with minimum clipping.


conorheins:
Thanks for the fix — agreed the asymmetry with `stable_xlogx` looks like an oversight from [799581a](https://github.com/infer-actively/pymdp/commit/799581a1ba76e5bc666a9e6b2ca015151f164380) (where both helpers were introduced together but only `stable_xlogx` got the `MINVAL` clip), and the only in-tree caller (`_state_prior_term`) optimizes `q`, not `p`, so clipping `y` won't introduce pathological gradients.

One nice-to-have (not a blocker): could you add a small regression test, something like:

```python
def test_stable_cross_entropy_handles_zero_in_y():
    x = jnp.array([0.5, 0.5])
    y = jnp.array([0.0, 1.0])
    assert jnp.isfinite(stable_cross_entropy(x, y))
```

That way we lock in the invariant and prevent regressions if the function gets touched again.

I've also kicked off the [Manual Branch Nightly](https://github.com/infer-actively/pymdp/actions/runs/24770443975) against this PR's head — happy to approve once that comes back green.

conorheins (approved):
Nightly passed -- approving

PR #396, opened 2026-04-22
ci: add release-please + PyPI trusted publishing

conorheins:
## Summary
- Adds release-please automation: on push to `main`, opens a "Release PR" bumping `pyproject.toml` + `setup.cfg` and updating `CHANGELOG.md`. Merging the release PR cuts a GitHub release + `vX.Y.Z` tag, then builds the distribution and publishes to PyPI via **trusted publishing (OIDC)** — no API tokens stored.
- Build + publish are jobs in the same workflow, gated on `releases_created == 'true'`, because `GITHUB_TOKEN`-created releases do not trigger downstream `release:published` workflows ([GitHub docs](https://docs.github.com/en/actions/concepts/security/github_token)). Consolidation avoids needing a PAT/App token.
- `setup.cfg` gets `# x-release-please-start-version` / `# x-release-please-end` block-marker annotations. (Inline `# x-release-please-version` would break `configparser`, which doesn't strip inline `#` comments and would read the version as literally `"1.0.0 # x-release-please-version"`.)

## Files
- `release-please-config.json` — `release-type: python`, tag format `vX.Y.Z`, changelog sections mapped from conventional commit types.
- `.release-please-manifest.json` — seeded at `1.0.0`.
- `.github/workflows/release-please.yml` — three jobs: `release-please` (always), then `build` and `publish` (only when a release is cut).
- `setup.cfg` — block-marker annotation around the version line.

## One-time GitHub setup needed before merge
1. **PyPI trusted publisher** — on pypi.org, add a pending publisher on the `inferactively-pymdp` project: owner `infer-actively`, repo `pymdp`, workflow `release-please.yml`, environment `pypi`.
2. **GitHub environment** — Settings → Environments → create `pypi` (optionally with a required reviewer for a manual gate on every publish).
3. **Enable workflow PR creation** — Settings → Actions → General → "Allow GitHub Actions to create and approve pull requests".

## Day-to-day flow after merge
Write commits in conventional-commit style (`feat:`, `fix:`, `deps:`, etc. — recent history already uses this loosely). Breaking changes: `feat!:` or a `BREAKING CHANGE:` footer → major bump. Release-please keeps a "chore: release X.Y.Z" PR open; merging it cuts a release and fires the PyPI publish in the same run.

## Test plan
- [x] Confirm `release-please` workflow runs on push to `main` (will show up in Actions tab after merge).
- [x] Verify the Release PR it opens bumps both `pyproject.toml` and `setup.cfg` and drafts `CHANGELOG.md`.
- [x] After the three one-time setup steps above, merge the Release PR and confirm:
  - [x] GitHub release + `vX.Y.Z` tag are created.
  - [x] `build` and `publish` jobs fire in the same workflow run.
  - [x] Publish successfully uploads to PyPI.

## Changes since first draft
Addressing review feedback from a second-opinion review:
- **Consolidated build+publish into release-please.yml**, gated on `releases_created`. Original `publish-pypi.yml` listening on `release:published` wouldn't have fired because `GITHUB_TOKEN`-created releases don't trigger downstream workflows.
- **Added `issues: write` permission**. Release-please applies `autorelease:pending` / `autorelease:tagged` labels to release PRs, which requires this scope.


conorheins:
Tagged some people to take a look at this @Arun-Niranjan @alec-tschantz @tverbele @dimarkov . Does anyone have experience with release-bots / release-please? I have limited experience with these so curious if others have opinions about how this should be best handled. I had Claude Code set this up quickly but had very little input into the design

dimarkov:
No idea what that is :) 


Sent with Proton Mail secure email.

Arun-Niranjan:
I have some experience with [goreleaser](https://goreleaser.com/getting-started/quick-start/) but not release please.

What you have here looks plausible at least - happy to approve if you'd like to test this out (but if you're not confident, you could always create a separate repo under the `inferactively` project and test this approach out there). 

One question - does `pyproject.toml` need some changes as well (like `setup.cfg`) ?

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,79 @@
...
+      tag_name: ${{ steps.release.outputs.tag_name }}
+    steps:
+      - uses: googleapis/release-please-action@v4
+        id: release
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          config-file: release-please-config.json
+          manifest-file: .release-please-manifest.json
+
+  build:
```
Arun-Niranjan, line 29: does this step need write permissions (or can you get away with read permissions only)? it will inherit everything from the default permissions you've defined in line 7

As a general principle, it's usually safer (if a bit more faff) to set default permissions to be as low as possible, and explicitly add more advanced permissions to jobs that need them

conorheins: good point thanks - now the workflow default is read permissions, but the release-please job elevates it to write persmissions. The `build` job thus inherits the read-only permission default and `publish` will override with write permsissions

conorheins:
> One question - does `pyproject.toml` need some changes as well (like `setup.cfg`) ?

it won't need changes actually, `release-please` will auto-detect / update `pyproject.toml`'s `[project] version` on its own, so we don't need to add anything there.

The block marker on `setup.cfg` is only because we registered it as an `extra-files` entry, so that the generic version updater knows it has to bump there as well. 


> happy to approve if you'd like to test this out

Sure let's approve/merge and then we can inspect the first release PR before completing the PyPi setup

Arun-Niranjan (approved):

PR #400, opened 2026-05-01
docs(notebooks): render math in mkdocs-jupyter notebooks

### LINKED ISSUE #386: Mathjax not rendering on MCTS + sophisticated inference on tmaze w/ distractors
In the Sophisticated inference notebook demo-ing performance of SI on the tmaze with distractors, the early markdown cells with mathjax are not rendering properly in the current documentation: https://pymdp-rtd.readthedocs.io/en/latest/tutorials/notebooks/examples/experimental/sophisticated_inference/mcts_generalized_tmaze/

conorheins:
## Summary

Fixes math rendering in Jupyter notebooks rendered through `mkdocs-jupyter` (e.g. [mcts_generalized_tmaze](https://pymdp-rtd.readthedocs.io/en/latest/tutorials/notebooks/examples/experimental/sophisticated_inference/mcts_generalized_tmaze/)). Previously `\begin{equation}` blocks, `$$…$$`, and `$…$` showed as raw LaTeX source instead of typeset math.

## Why it broke

`mkdocs-jupyter` runs notebook cells through `nbconvert`, which leaves raw `$…$` / `$$…$$` / `\begin{…}` in the HTML output inside `.jp-Notebook` (and inside generic descendant `<div>`s like `.jp-RenderedMarkdown`, `<p>`, etc.). It does **not** wrap each math expression in `.arithmatex` the way `pymdownx.arithmatex` does for regular markdown.

The previous `docs-mkdocs/javascripts/mathjax.js` was tuned for `pymdownx.arithmatex` only:
- `inlineMath: [["\\(", "\\)"]]` / `displayMath: [["\\[", "\\]"]]` — no dollar-sign delimiters
- `ignoreHtmlClass: ".*|"` + `processHtmlClass: "arithmatex"` — ignore the entire DOM **except** `.arithmatex` containers

Because notebook math sits inside generic divs that don't carry an `.arithmatex` class, MathJax saw nothing to process.

## Fix

`docs-mkdocs/javascripts/mathjax.js`:
- Add `["$", "$"]` and `["$$", "$$"]` to `inlineMath` / `displayMath` so the dollar-sign delimiters that nbconvert preserves are recognised. (`processEscapes: true` is already set, so `\$` still escapes a literal dollar.)
- Add `jp-Notebook` to `processHtmlClass`, alongside the existing `arithmatex`.
- Replace the aggressive `ignoreHtmlClass: ".*|"` with the MathJax default `tex2jax_ignore`. `processHtmlClass` resets MathJax's ignore state when matched, but the descent re-evaluates each child class and the `.*|` regex matched everything in between (`.jp-Cell`, `.jp-RenderedMarkdown`, `<p>`, …) — re-triggering ignore before MathJax could reach the actual math text. Switching to the default lets MathJax descend through those generic wrappers, while `processHtmlClass` still keeps math processing scoped to `.arithmatex` and `.jp-Notebook` subtrees.

Code blocks are unaffected — `<pre>` / `<code>` are in `skipHtmlTags` by default, so dollar signs in shell snippets aren't parsed as math.

## Test plan

Verified locally with `mkdocs serve` + headless Chrome (puppeteer):

| Page | mjx-containers | mjx-merror |
|---|---|---|
| `mcts_generalized_tmaze` | 24 | 0 |
| `inference_methods_comparison` | 1 | 0 |
| `guides/generative-model-structure` (markdown w/ arithmatex) | 8 | 0 |
| `tmaze_demo`, `mcts_graph_world`, `model_construction_tutorial` (no math) | 0 | 0 |
| `index`, `api/agent`, `installation` | 0 | 0 |

Visual spot-check of the rendered notebook confirms `\begin{equation}` blocks, display `$$…$$`, and inline `$…$` all typeset correctly, with no leftover raw LaTeX.

- [x] Local mkdocs build succeeds
- [x] Math renders on the affected notebook
- [x] Math still renders on regular markdown pages (no regression for `pymdownx.arithmatex`)
- [x] Notebooks without math show 0 MathJax errors
- [x] Verify on the deployed RTD build once this lands

Fixes #386

### DIFF
--- a/docs-mkdocs/javascripts/mathjax.js
+++ b/docs-mkdocs/javascripts/mathjax.js
@@ -1,7 +1,13 @@
+// Dollar-sign delimiters and `jp-Notebook` are needed for math in Jupyter
+// notebooks: mkdocs-jupyter renders cells via nbconvert, which leaves raw
+// `$...$` / `$$...$$` / `\begin{...}` in the HTML inside `.jp-Notebook`
+// rather than wrapping them in `.arithmatex` spans the way pymdownx does.
+// `ignoreHtmlClass` stays at the default so MathJax can descend through the
+// generic divs nested between `.jp-Notebook` and the actual math text.
 window.MathJax = {
   tex: {
-    inlineMath: [["\\(", "\\)"]],
-    displayMath: [["\\[", "\\]"]],
+    inlineMath: [["\\(", "\\)"], ["$", "$"]],
+    displayMath: [["\\[", "\\]"], ["$$", "$$"]],
     processEscapes: true,
     processEnvironments: true,
   },
@@ -9,8 +15,8 @@ window.MathJax = {
     fontCache: "global",
   },
   options: {
-    ignoreHtmlClass: ".*|",
-    processHtmlClass: "arithmatex",
+    ignoreHtmlClass: "tex2jax_ignore",
+    processHtmlClass: "arithmatex|jp-Notebook",
   },
 };
 


conorheins:
RTD build looks good, checking off last task on the list

PR #406, opened 2026-05-25
docs(notebooks): add google colab badge and install cell to model construction notebook

Arun-Niranjan:
It would be good to be able to run our notebooks in google colab (makes talks and live demos much easier).

This change modifies the `model_construction_tutorial` notebook so that this works.

I'll do the others in a follow-up PR, just wanted this one to be small to make the changes obvious

### DIFF
--- a/examples/api/model_construction_tutorial.ipynb
+++ b/examples/api/model_construction_tutorial.ipynb
@@ -36,7 +36,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 33,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -47,7 +47,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 34,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -62,7 +62,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 35,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -101,7 +101,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 36,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -140,7 +140,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 37,
+   "execution_count": 1,
    "metadata": {},
    "outputs": [
     {
@@ -194,7 +194,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 38,
+   "execution_count": 2,
    "metadata": {},
    "outputs": [
     {
@@ -261,7 +261,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 39,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


examples/api/model_construction_tutorial.ipynb
```diff
@@ -29,7 +36,18 @@
...
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import sys\n",
+    "if \"google.colab\" in sys.modules:\n",
+    "    %pip install \"inferactively-pymdp\" -q"
```
Arun-Niranjan, line 45: logic:
- we only run this install if we are in a colab space
- we can install the vanilla dependencies, as colab ships with JAX and CUDA support already there. (in fact if you try `%pip install "inferactively-pymdp[gpu]" -q` you get an error)

Arun-Niranjan:
@conorheins I can't remember what we decided on notebook outputs? it does make the diffs much more painful to evaluate, but I think for tutorial notebooks especially they're worth keeping so that people can see what the expected results are without necessarily having to run it.

conorheins:
This is a nice idea and agree will make quick experimentation for users and demo-ing during presentations much more frictionless.

> @conorheins I can't remember what we decided on notebook outputs? it does make the diffs much more painful to evaluate, but I think for tutorial notebooks especially they're worth keeping so that people can see what the expected results are without necessarily having to run it.

good question -- yeah the policy was to keep saved outputs. the relevant prior work are PRs #367 and #375. Short version of the policy:
- every non-legacy notebook under examples/ lives in either `test_notebooks/ci_notebooks.txt`(PR-CI tier) or `tests/notebooks/nightly_notebooks.txt` (heavier nightly tier).
- both tiers keep saved outputs -- exactly for the reason you said
- to keep diffs tractable, a local precommit hook (`scripts/notebook_precommit.py`) does two things on commit: strips top-level `kernelspec` / `language_info` metadata, and standardizes execution counts (output-bearing cells get `1, 2, ..., N` and empty-output cells get `null`). That way local kernel history doesn't get committed

So keeping outputs in the `model_construction_tutorial` as you did here is the right call. The one thing missing is running the hooks -- I pulled the branch down and the current state still has `kernelspec` / `language_info` populated and execution counts in the 33–48 range from your local kernel. If you run:

```bash
uv sync --group dev
uv run --group dev pre-commit install
uv run --group dev pre-commit run --all-files
```

it'll fix the metadata + renumber the execution counts, and the diff will get a lot easier to review. Worth installing before you push [#407](https://github.com/infer-actively/pymdp/pull/407) too since it touches all 20 notebooks. Docs for all of this are in `test/notebooks/README.md` and the "Recommended local hooks" section of [`CONTRIBUTING.md`](https://github.com/infer-actively/pymdp/blob/main/CONTRIBUTING.md#recommended-local-hooks).

FYI the hooks aren't wired into a CI job ( only the ruff linting runs there) so this won't block the merge, but it's the documented convention.

Arun-Niranjan:
ok that's done for this one, once this is merged I'll update the PR stacked on top.

conorheins (approved):
LGTM!

One heads-up for #407: most of the other 19 notebooks haven't been touched since the notebook standardization rule landed in #375, so when you run the hooks on that stacked PR you'll see each notebook diff come out bigger than just "add 2 cells"; there'll be incidental kernelspec/language_info stripping and exec-count renumbering baked in alongside the badge + install cell. That's expected and fine, just don't be surprised by the diff sizes

PR #407, opened 2026-05-25
docs(notebooks): add colab badge and install cell to other notebooks

Arun-Niranjan:
Follow on from https://github.com/infer-actively/pymdp/pull/406

Diff is a bit larger than expected because of https://github.com/infer-actively/pymdp/pull/406#pullrequestreview-4362744170


conorheins (approved):
Thanks for this, LGTM!

PR #409, opened 2026-05-25
docs: remove legacy docs folder

### LINKED ISSUE #408: Can we remove the docs folder?
The old docs folder is still available at https://github.com/infer-actively/pymdp/tree/main/docs - I believe this is entirely superseded by the autogenerated docs in https://github.com/infer-actively/pymdp/tree/main/docs-mkdocs

This is confusing for new users, who reasonably from a search engine go straight to the docs folder, which is out of date.

We should remove this docs folder entirely now we have completed the v1 release.

Arun-Niranjan:
Addresses: 
- https://github.com/infer-actively/pymdp/issues/408

and also:
- https://github.com/infer-actively/pymdp/security/dependabot/1
- https://github.com/infer-actively/pymdp/security/dependabot/2
- https://github.com/infer-actively/pymdp/security/dependabot/4
- https://github.com/infer-actively/pymdp/security/dependabot/5
- https://github.com/infer-actively/pymdp/security/dependabot/6

because we don't use jinja when generating docs the new way


conorheins (changes requested):
thanks for this, agree with the motivation in #408 and getting the 5 jinja2 dependabot alerts to disappear as a side effect is a nice bonus.

Two small things before we land:

**1. Preserve three orphan notebooks.** these have no counterpart anywhere in `examples/`:
- `docs/notebooks/active_inference_from_scratch.ipynb`
- `docs/notebooks/pymdp_fundamentals.ipynb`
- `docs/notebooks/using_the_agent_class.ipynb`

could you move them into `examples/legacy/` before deleting the rest? i think they are useful educational material despite being outdated in terms of referencing numpy backend. that dir is already excluded from the notebook manifests so there won't be any CI overhead. the other three notebooks in `docs/notebooks/` (`cue_chaining_demo`, `tmaze_demo`, `free_energy_calculation`) already have counterparts in examples/ so those are fine to drop.

**2. stale line to delete.** `docs-mkdocs/development/viewing-docs.md:34` says `` `docs/` is retained as legacy Sphinx-era content for compatibility/history.`` which becomes false after this PR, so worth dropping in the same go.

once those tiny changes are in, think this PR is good to go

conorheins (approved):
great, lgtm!

PR #414, opened 2026-05-31
docs: add mnist rgm notebook

Arun-Niranjan:
Initial attempt post v1 release to implement scale free active inference MNIST classifier as per this [paper](https://www.frontiersin.org/journals/network-physiology/articles/10.3389/fnetp.2025.1521963/full) and this [demo script in SPM](https://github.com/spm/spm/blob/main/toolbox/DEM/DEM_MNIST_RGM.m). 

I could not achieve the accuracy levels of ~95% mentioned in the paper (this implementation gets to ~87% (edit: after fixing two bugs with learning order and soft likelihood messages, this is now up to ~90%) in the test set), and there are quite a few deviations between the paper and the demo script, or underspecified methods in general.

Of particular note:
- the SPM demo additionally includes asymptotic forgetting controlled with an `eta` hyperparameter
- the SPM demo does joint optimisation of A and Q during parameter learning with expectation maximisation, which we don't have in pymdp - we have the infer_parameters(...) method with a pre-specified learning rate. (edit: I've added a custom implementation of this in `_interleaved_em(...)`)

Because this PR required quite a lot of extra infrastructure to get going, I recommend checking out the branch,  running the `mnist_classifier.ipynb` notebook and then clicking back into the various functions to get a better idea of the implementation

Disclaimer: The work on this branch was done with heavy use of Claude code, especially for trying to compare against the SPM demo and also to make siginificant performance improvements with JAX optimisations

THIS PR SHOULD NOT BE MERGED AS IT IS. I would like to get some broad feedback on direction and algorithmic choices, and once we can get decent accuracy we can decide on:
- how much of this infrastructure (if any) is worth moving into pymdp core package
- how much should live in code next to the notebook to demonstrate to users how they can do scale free active inference in this way
- how to extend to the dove animation or atari games (I have simply not considered actions yet)
- how to have at least some of this notebook tested (it's currently not suitable for CI because it will need the MNIST dataset to run anything useful)

Then I will split code from this branch into sensible chunks that can be properly reviewed.


conorheins:
Thanks for this @Arun-Niranjan and sorry for the late review.

I pulled this down and ran the training path on the full 10k images. I got roughly 83% on the first 1,000 official test images and 86.3% on the SPM-style `training[10000:11000]` slice. Structure learning and generation also work, so I think this is a useful and promising direction.

My main thoughts would be: maybe don't spend the next phase tuning performance to 95.1% quite yet. The SPM reference fits its SVD bases and bin ranges on all 11,000 images before splitting 10k/1k, and its “test” set is drawn from MNIST’s training split. Your implementation fits the basis from only the 130 structure exemplars and evaluates the official test set, so it's not entirely surprising that the reported accuracies are not comparable.

I’d suggest defining two modes:

1. An exact SPM-parity mode, including the original protocol, for validating the port.
2. A clean benchmark mode with train-only preprocessing and the complete official test set.

Before further optimisation, I think we should validate several algorithmic details against small exported SPM fixtures:

- soft hierarchical likelihood messages: the current pymdp categorical-observation calculation differs from SPM for non-one-hot beliefs (in pymdp the log happens inside the expectation $ \sum_o q(o) \log p(o|s)$ , whereas in SPM it happens outside i.e.$ \log \sum_o q(o) p(o|s) $
- the mutual-information-based gate, which SPM applies **per** likelihood mapping rather than once across a whole level like you do here;
- full-spectrum SVD thresholding followed by the 16-mode truncation;
- symmetric zero-centred bins;
- explicit masking of padded child outcomes;
- total hierarchical ELBO, so you can reproduce the paper’s accuracy-versus-coverage result, as done [here](https://github.com/spm/spm/blob/23a9359a60bb13eec207591e238ec24b367f42ad/toolbox/DEM/DEM_MNIST_RGM.m#L418-L450).

I’d keep the implementation under `examples/experimental` for now and add small deterministic unit/golden tests around those primitives. The expensive notebook can be a nightly/manual integration test. Once parity is demonstrated, we can decide which pieces -- hierarchical message passing, conjugate updates, and the MI gate -- are mature enough to promote into pymdp proper.

Actions or the Dove/Atari extensions do not need to be in scope for this PR; establishing a correct static MNIST RGM is already a substantial and useful milestone.

examples/renormalising_generative_models/mutual_information.py
```diff
@@ -0,0 +1,57 @@
...
+
+from pymdp.agent import Agent
+
+
+# ---------------------------------------------------------------------------
+# Mutual information helpers
+# ---------------------------------------------------------------------------
+
+
+def _mi_from_pA(pA_list: list[jnp.ndarray]) -> jnp.ndarray:
```
conorheins: from what i can tell, this is summing across every patch and modality, whereas [SPM gates each likelihood mapping independently](https://github.com/spm/spm/blob/23a9359a60bb13eec207591e238ec24b367f42ad/toolbox/DEM/spm_MDP_VB_XXX.m#L1450-L1464). With $\beta = 512$, this could materially change what gets learned

examples/renormalising_generative_models/discretise.py
```diff
@@ -0,0 +1,308 @@
...
+    centred_packed = jnp.concatenate([centred, L_row], axis=1)  # (ng*ng, N+1, n_pixels)
+
+    # SVD returns at most min(N, n_pixels) components; pad to k
+    max_rank = min(N, n_pixels)
+
+    def svd_one(data_packed):
+        data = data_packed[:N, :]   # (N, n_pixels)
+        L = data_packed[N, 0]       # scalar L for this tile
+
+        _, S_full, Vh = jnp.linalg.svd(data, full_matrices=False)
```
conorheins, line 162: it looks like you first truncate to the top `nk`singular values here before calculating the 'retention denominator', whereas SPM thresholds using the full spectrum (more than nk vlaues) and then only caps them post-threshold at 16: https://github.com/spm/spm/blob/23a9359a60bb13eec207591e238ec24b367f42ad/spm_svd.m#L35-L82

examples/renormalising_generative_models/discretise.py
```diff
@@ -0,0 +1,308 @@
...
+
+    # Project to compute bin centres
+    # variates: for each group, project centred data onto V
+    # centred: (ng*ng, N, n_pixels), V_flat: (ng*ng, n_pixels, k)
+    variates_flat = jnp.einsum('gnf,gfk->gnk', centred, V_flat)  # (ng*ng, N, k)
+    variates = variates_flat.reshape(ng, ng, N, k)
+
+    mode_mask = (S > 0).astype(jnp.float32)
+
+    def symmetric_bin_centres(v):
```
conorheins, line 207: [SPM uses `[-max(abs(u)), +max(abs(u))]` ](https://github.com/spm/spm/blob/23a9359a60bb13eec207591e238ec24b367f42ad/toolbox/DEM/spm_rgb2O.m#L127-L155)

examples/renormalising_generative_models/rgm_hierarchy.py
```diff
@@ -0,0 +1,513 @@
...
+        levels: list[RGMLevel],
+        n_classes: int = 10,
+    ):
+        self.config = config
+        self.basis = basis
+        self.y_exemplars = np.asarray(y_exemplars)
+        self.levels = levels
+        self.n_classes = n_classes
+
+    @classmethod
```
conorheins, line 89: maybe i'm confused here, but doesn't this conflate the fitting of the SVD frnot-end with the fitting of the structuer learning algorithm? I thought they should be separate sets of inputs --one dataset for fitting the front end and then 13-per-class exemplars for fast structure learning itself, using the separately-learned SVD basis frontend

examples/renormalising_generative_models/hierarchical.py
```diff
@@ -0,0 +1,331 @@
...
+        (agent, valid_states_mask) where valid_states_mask is (n_patches, max_states)
+    """
+    n_grid = stats.num_states.shape[0]
+    n_patches = n_grid * n_grid
+    n_mod = 4  # always 4 modalities (2x2 children)
+
+    num_states_flat = stats.num_states.flatten()  # (n_patches,)
+    max_states = int(num_states_flat.max())
+
+    # Max child states across entire child grid — observation dimension
```
conorheins, line 73: The code from here seems to suggest that probability can leak into nonexistent states.
The batched JAX representation pads every local mapping to the largest child-state vocabulary. Concentration is then added to those padded rows, while only hidden-state masks (not observation mask) are consistently carried through inference and generation, see also https://github.com/infer-actively/pymdp/blob/26184495d37e55e14b7a84eae0344a92a0604705/examples/renormalising_generative_models/rgm_hierarchy.py#L370-L385

examples/renormalising_generative_models/training.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def _interleaved_em(
+    pA_list: list[jnp.ndarray],
+    obs_soft_list: list[jnp.ndarray],
+    D: jnp.ndarray,
+    valid_mask: jnp.ndarray,
+    num_iter: int = 16,
+) -> tuple[list[jnp.ndarray], jnp.ndarray]:
+    """Interleaved Q-A EM matching MATLAB's spm_VBX within-level solver.
```
conorheins, line 99: i don't think this matches the code path the demo actually runs: the 16-iteration `qa = pa; qa += cross(O,Q)` scheme lives in [`spm_backwards`](https://github.com/spm/spm/blob/23a9359a60bb13eec207591e238ec24b367f42ad/toolbox/DEM/spm_MDP_VB_XXX.m#L1991-L2022), which is only entered [when `OPTIONS.B=1`](https://github.com/spm/spm/blob/23a9359a60bb13eec207591e238ec24b367f42ad/toolbox/DEM/spm_MDP_VB_XXX.m#L1421) -- and the MNIST demo [sets `OPTIONS.B = 0`](https://github.com/spm/spm/blob/23a9359a60bb13eec207591e238ec24b367f42ad/toolbox/DEM/DEM_MNIST_RGM.m#L315). `spm_VBX` itself is [explicitly non-iterative](https://github.com/spm/spm/blob/23a9359a60bb13eec207591e238ec24b367f42ad/toolbox/DEM/spm_VBX.m#L13-L23) (one belief-propagation pass with A held fixed), so the demo's update is a single inference under the *prior* A followed by a single Dirichlet accumulation. iterating Q against an A refit to the same image is self-reinforcing and will sharpen the counts towards one-hot relative to SPM -- worth trying `num_iter=1` and comparing before attributing the accuracy gap elsewhere

examples/renormalising_generative_models/training.py
```diff
@@ -0,0 +1,233 @@
...
+    # --- Top-down pass: refine beliefs using cls posterior ---
+    refined_soft = _top_down_refinement_pass(
+        levels, valid_masks, cls_agent, q_cls, level_soft_beliefs, level_obs_lists
+    )
+
+    # --- Interleaved EM updates at every hierarchical level (matching spm_VBX) ---
+    new_levels = []
+    for lv_idx in range(n_hier):
+        level = levels[lv_idx]
+        D_em = refined_soft[lv_idx]   # top-down refined prior over states
```
conorheins: `refined_soft[lv_idx]` is a posterior that has already absorbed this level's observations (the refinement pass re-ran `infer_states` on the same obs list), and the E-step in `_interleaved_em` then multiplies the same observations in again -- so the evidence gets counted twice. in SPM the prior for the learning posterior is the top-down message alone ([`mdp.D{f} = spm_dot(A, Q_parent)`](https://github.com/spm/spm/blob/23a9359a60bb13eec207591e238ec24b367f42ad/toolbox/DEM/spm_MDP_VB_XXX.m#L1000) folded into the child's D), with the observation entering once. i think passing the `_top_down_D_hierarchical` / `_top_down_D_from_cls` output here instead of the refined posterior would be closer to the reference

examples/renormalising_generative_models/training.py
```diff
@@ -0,0 +1,233 @@
...
+        if lv_idx == 0:
+            n_obs_l1 = level.pA[0].shape[1]  # n_levels (SVD bin outcomes)
+            obs_soft = [jax.nn.one_hot(o, n_obs_l1) for o in level_obs_lists[0]]
+        else:
+            obs_soft = level_obs_lists[lv_idx]  # already (n_patches, max_child_states)
```
conorheins: the outcome side of the `cross(O, Q)` accumulation here (and `obs_soft_cls = [child_beliefs]` below for the classification level) comes from the *unsupervised* bottom-up pass. in SPM the supervised one-hot D conditions the whole downward sweep before inference happens, so the child posteriors that come back up as the outcomes O are label-conditioned at every level by the time `qa += cross(O,Q)` runs. early in training this matters: SPM's counts get pulled toward the correct class's exemplar states, whereas these reinforce whatever the unsupervised parse said (only the Q side is label-corrected). using the refined (post-top-down) beliefs as the observation side would be closer to the reference

examples/renormalising_generative_models/rgm_hierarchy.py
Arun-Niranjan: Hmm I thought the SVD frontend _was_ the fast structure learning with 13 exemplars - and then we use the next 10k images for parameter learning on top of the one-hot encodings.

I've made this change anyway and it seems to perform better, so the confusion must be on my end

examples/renormalising_generative_models/training.py
Arun-Niranjan: This was the biggest source of inaccuracy - great spot!

Arun-Niranjan:
Currently stacked on top of https://github.com/infer-actively/pymdp/pull/441 which sorts out linting

conorheins (changes requested):
Thanks, Arun. This is a substantial improvement, and several of the original review points now look properly addressed: the full-spectrum SVD threshold, symmetric bins, per-mapping MI gate, separation of basis/structure inputs, top-down priors, and label-conditioned child outcomes.

I also ran the revised implementation. The new tests pass, and a full clean run reached 92.44% on the MNIST test set (10k images), so the classifier is now producing an encouraging result.

I don’t think we can call the SPM-parity work complete yet, though. The main remaining issue is the learning order: with `num_iter=1`, the code first forms a sample-updated likelihood from `O ⊗ D`, infers `Q` under that likelihood, and then forms the final counts. `OPTIONS.B=0` in SPM instead infers under the fixed prior-normalised `A` and adds `O ⊗ Q` only afterward. The new manual test currently duplicates the implementation’s update order, so it does not establish parity.

The soft-message likelihood also still differs: the generic pymdp inference path uses `Σ q(o) log A(o|s)`, while `spm_VBX` uses `log(Σ A(o|s)q(o))`. Because the upper-level messages are soft, this is a substantive difference rather than a notation issue. The overall bottom-up/top-down schedule also remains an approximation to SPM’s recursive schedule.

For the next iteration, if we think 100% parity is important, I’d suggest focusing narrowly on:

- implementing a dedicated SPM-style RGM inference primitive;
- making the parity update explicitly “infer using prior A, then accumulate once”;
- adding a few tiny MATLAB-exported golden fixtures for the soft likelihood, posterior, count update and MI gate;
- moving or exposing the tests under `test/` so CI actually runs them;
- masking padded hidden-state columns as well as outcome rows;
- adding ELBO/risk–coverage evaluation if reproducing the paper’s selective-classification result remains a goal.

It would also help to distinguish “SPM data protocol” from “SPM algorithmic parity” in the strategy naming. If exact parity is not the intended goal, I’m comfortable treating this as a promising experimental RGM-inspired classifier, provided the documented deviations are explicit.

PR #415, opened 2026-07-04
fix(maths): eliminate cancellation error in exact parameter info-gain weight

### LINKED ISSUE #340: Numerical instability in `_exact_wnorm` / parameter information gain for small Dirichlet counts (MINVAL sensitivity, EFE oscillations)
### Summary
A user reported **bumpy / oscillatory minima in EFE** and degraded control performance on `v1.0.0_alpha` when `use_param_info_gain=True` and `MINVAL = jnp.finfo(float).eps`. Setting `MINVAL = 1e-8` reduced oscillations and improved performance. This points to a **numerical stability** issue, likely in the parameter information gain path (`calc_pA_info_gain` / `calc_pB_info_gain`).

### Suspected cause
`_exact_wnorm` currently computes terms including `1/A` and `digamma(A)` for Dirichlet counts `A`. For very small concentrations, `1/A` and `digamma(A)` become huge with opposite signs and should cancel (since `digamma(x) ~ -1/x` as `x→0`). In finite precision (esp. float32/GPU), this cancellation can be unstable, causing large fluctuations in `wnorm` → `param_info_gain` → EFE.

Raising `MINVAL` reduces intermediate magnitudes, but changes the effective model, so I prefer a principled fix.

### Proposed fix
Rewrite `_exact_wnorm` to eliminate explicit `1/x` terms using:

$$\psi(x+1) = \psi(x) + \frac{1}{x} $$

so that `1/A + digamma(A)` becomes `digamma(A+1)`. Equivalent but more numerically stable:

$$w = \log(\alpha_0) - \log(\alpha) + \psi(\alpha+1) - \psi(\alpha_0+1) $$

(up to the historical minus sign in the current implementation).

### Tests to add
- `_exact_wnorm` matches the stable Dirichlet-KL closed form above (within tolerance).
- “Small symmetric concentration” stress-case (uniform tiny counts) stays **O(1)** (≈ `log(K)`), not O(1/t), and gradients are finite.
- `calc_pA_info_gain` matches a direct mean-field reference computed from `_exact_wnorm`.

### Links
- Related: #336 (distributed obs likelihood issue)
- Fix applied: #337
- Related: #191 (replacing approximate spm_wnorm with exact info gain)
- Fix applied: #193

gauraaansh:
Fixes #340

`_exact_wnorm` computes eq. (D.15) from Da Costa et al. 2020 as:

    w = log(sumA) - log(A) + 1/A - 1/sumA + digamma(A) - digamma(sumA)

As A shrinks toward MINVAL, `1/A` and `digamma(A)` both diverge with opposite
sign and are supposed to cancel down to an O(1) value. In float32 that
cancellation is inexact, so the term comes out wrong by several orders of
magnitude for small Dirichlet counts. The error is also erratic rather than
a smooth degradation, matching the oscillating behavior described in the
issue.

Using the digamma recurrence `digamma(x+1) = digamma(x) + 1/x`, substitute
for `1/A` and `1/sumA`:

    w = log(sumA) - log(A)
        + [digamma(A+1) - digamma(A)]
        - [digamma(sumA+1) - digamma(sumA)]
        + digamma(A) - digamma(sumA)

The `digamma(A)` and `digamma(sumA)` terms cancel exactly, leaving:

    w = log(sumA) - log(A) + digamma(A+1) - digamma(sumA+1)

This reformulation is algebraically identical to the original expression. It
changes only the numerical evaluation order. The divergent terms are removed
before any floating-point evaluation happens, so the implementation no
longer depends on cancellation between divergent quantities.

## Testing

- Randomized stress testing against float64 and mpmath, plus a deterministic
  logarithmic sweep from 1e-20 to 1e3: the original formula's error grows to
  roughly six orders of magnitude near the clipping threshold, while the
  rewritten version stays within float32's ordinary rounding error across the
  same range.
- Verified under jit, vmap, and autodiff (the composition control.py and
  agent.py use for policy inference), including the zero-entry case from
  Bayesian model reduction.
- The existing correctness test previously compared the implementation to a
  hand-copied version of the same pre-fix formula, so it would pass
  regardless of correctness. Added a test against `-log(K)`, the closed-form
  limit eq. (D.15) converges to for a symmetric Dirichlet as concentration
  goes to zero. This reference is independent of either formula.
- Full suite passes (294 passed, 1 skipped, one pre-existing unrelated
  failure from a missing `numpyro` dependency). `ruff check` is clean.

### DIFF
--- a/test/test_param_info_gain_jax.py
+++ b/test/test_param_info_gain_jax.py
@@ -49,7 +49,9 @@ def test_exact_wnorm_mathematical_correctness():
     )
     expected = -expected  # minus sign in the implementation
 
-    np.testing.assert_allclose(result, expected, rtol=1e-4, atol=1e-4)
+    # Measured max relative diff between the two forms on this input is ~4.8e-6 in
+    # float32 (see discussion on PR #415); rtol/atol below leave a ~3-11x margin above that.
+    np.testing.assert_allclose(result, expected, rtol=1e-5, atol=1e-6)
 
     # Verify output shape matches input
     assert result.shape == A.shape
@@ -61,9 +63,9 @@ def test_exact_wnorm_symmetric_tiny_concentrations_match_log_k(K, a):
     """For a symmetric Dirichlet (all K entries equal to `a`), the exact value of eq. (D.15)
     as `a -> 0` is `-log(K)`: both `digamma(a+1)` and `digamma(K*a+1)` tend to `digamma(1)`
     and cancel, leaving only `log(K)`. This is an independent closed-form reference (no
-    digamma evaluation needed at the limit), so it directly catches the cancellation bug:
-    the pre-fix direct-expansion formula drifts further from `-log(K)` as `a` shrinks
-    instead of converging to it.
+    digamma evaluation needed at the limit), so it directly catches the cancellation error
+    described in #340: the pre-fix direct-expansion formula drifts further from `-log(K)`
+    as `a` shrinks instead of converging to it.
     """
     A = jnp.full((K, 1), a)
     result = _exact_wnorm(A)
--- a/test/test_param_info_gain_jax.py
+++ b/test/test_param_info_gain_jax.py
@@ -32,24 +32,43 @@ def test_exact_wnorm_mathematical_correctness():
     # Simple 2x2 case with known values
     A = jnp.array([[1.0, 2.0],
                    [3.0, 4.0]])
-    
+
     result = _exact_wnorm(A)
-    
-    # Manually compute expected result using the formula
-    # wA = log(A.sum(0)) - log(A) + 1/A - 1/A.sum(0) + digamma(A) - digamma(A.sum(0))
+
+    # Reference: direct expansion of eq. (D.15), stable at this magnitude since A is
+    # far from MINVAL. Mathematically equivalent to the implementation (see the digamma
+    # recurrence note in _exact_wnorm's docstring), so both should agree up to float32
+    # rounding -- the two forms don't round identically since they go through different
+    # sequences of operations.
     A_sum = A.sum(axis=0)  # [4.0, 6.0]
-    
+
     expected = (
         jnp.log(A_sum) - jnp.log(A) +
         1.0 / A - 1.0 / A_sum +
         jnp.array(digamma(A)) - jnp.array(digamma(A_sum))
     )
     expected = -expected  # minus sign in the implementation
-    
-    np.testing.assert_allclose(result, expected, rtol=1e-6, atol=1e-8)
-    
+
+    np.testing.assert_allclose(result, expected, rtol=1e-4, atol=1e-4)
+
     # Verify output shape matches input
-    assert result.shape == A.shape 
+    assert result.shape == A.shape
+
+
+@pytest.mark.parametrize("K", [2, 5, 10, 50])
+@pytest.mark.parametrize("a", [1e-8, 1e-12, 1e-15])
+def test_exact_wnorm_symmetric_tiny_concentrations_match_log_k(K, a):
+    """For a symmetric Dirichlet (all K entries equal to `a`), the exact value of eq. (D.15)
+    as `a -> 0` is `-log(K)`: both `digamma(a+1)` and `digamma(K*a+1)` tend to `digamma(1)`
+    and cancel, leaving only `log(K)`. This is an independent closed-form reference (no
+    digamma evaluation needed at the limit), so it directly catches the cancellation bug:
+    the pre-fix direct-expansion formula drifts further from `-log(K)` as `a` shrinks
+    instead of converging to it.
+    """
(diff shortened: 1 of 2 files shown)


conorheins:
weirdly it appears the 3.14 unit tests are timing out on the `infer_states` optimization  notebooks, and the other 3 tested versions (3.11-3.13) are just being cancelled. will look into this when i have time

conorheins:
> weirdly it appears the 3.14 unit tests are timing out on the `infer_states` optimization notebooks, and the other 3 tested versions (3.11-3.13) are just being cancelled. will look into this when i have time

opened #417 to address this, let's land that first and then we can unblock your two PRs from merging

conorheins:
> > weirdly it appears the 3.14 unit tests are timing out on the `infer_states` optimization notebooks, and the other 3 tested versions (3.11-3.13) are just being cancelled. will look into this when i have time
> 
> opened #417 to address this, let's land that first and then we can unblock your two PRs from merging

Merged #417 -- @gauraaansh can you please rebase this branch to main and then we can retrigger CI?

test/test_param_info_gain_jax.py
```diff
@@ -32,24 +32,43 @@ def test_exact_wnorm_mathematical_correctness():
...
+    assert result.shape == A.shape
+
+
+@pytest.mark.parametrize("K", [2, 5, 10, 50])
+@pytest.mark.parametrize("a", [1e-8, 1e-12, 1e-15])
+def test_exact_wnorm_symmetric_tiny_concentrations_match_log_k(K, a):
+    """For a symmetric Dirichlet (all K entries equal to `a`), the exact value of eq. (D.15)
+    as `a -> 0` is `-log(K)`: both `digamma(a+1)` and `digamma(K*a+1)` tend to `digamma(1)`
+    and cancel, leaving only `log(K)`. This is an independent closed-form reference (no
+    digamma evaluation needed at the limit), so it directly catches the cancellation bug:
```
conorheins: maybe remove this phrase 'the cancellation bug' since that's referring to an issue as if it's known / in the common ground of the user who is reading this, you could reference the issue #340 if you want though

test/test_param_info_gain_jax.py
```diff
@@ -32,24 +32,43 @@ def test_exact_wnorm_mathematical_correctness():
...
+    the pre-fix direct-expansion formula drifts further from `-log(K)` as `a` shrinks
+    instead of converging to it.
+    """
+    A = jnp.full((K, 1), a)
+    result = _exact_wnorm(A)
+    expected = -jnp.log(float(K))
+    np.testing.assert_allclose(result, expected, rtol=1e-4, atol=1e-4)
```
conorheins: this is about 20x slacker than the measured need (4.8e-6), so maybe try to do it with rtol=1e-5, atol=1-6 instead? it is defensible either way though, since digamma accuracy varies acros XLA backends

test/test_param_info_gain_jax.py
```diff
@@ -32,24 +32,43 @@ def test_exact_wnorm_mathematical_correctness():
...
         jnp.log(A_sum) - jnp.log(A) +
         1.0 / A - 1.0 / A_sum +
         jnp.array(digamma(A)) - jnp.array(digamma(A_sum))
     )
     expected = -expected  # minus sign in the implementation
-    
-    np.testing.assert_allclose(result, expected, rtol=1e-6, atol=1e-8)
-    
+
+    np.testing.assert_allclose(result, expected, rtol=1e-4, atol=1e-4)
```
conorheins: why did you change the rtol and atol here? is it because of this sentence from the comment?

"the two forms don't round identically since they go through different
    # sequences of operations."?

conorheins (commented):
Looks good overall, just a few comments/questions

test/test_param_info_gain_jax.py
gauraaansh: My apologies, will reference #340 here.

test/test_param_info_gain_jax.py
gauraaansh: Yes,  before this edit , the test's expected was the direct-expansion formula _exact_wnorm used internally, so result and expected were two evaluations of an identical op sequence, and the rtol=1e-6/atol=1e-8 was just absorbing JAX level nondeterminism. 

The implementation moved to the digamma-recurrence form, I did a  test where I kept the old direct-expansion formula as reference. The two are algebraically equivalent but operationally different and so they round differently in float32, and that's why we get  ~4.8e-6 discrepancy thus loosened the tolerance.

from what i measured : 
max relative diff: 4.81e-6
max absolute diff: 7.15e-7

gauraaansh: Just saw the other comment, making these values rtol=1e-5, atol=1e-6 instead, we still have good margins (3.5x to 11.6x from a very short test)

conorheins (approved):

Issue #227, opened 2025-08-22
Agent `policies` field in equinox module has no batch_dim


tverbele:
The `Agent` Equinox module has a `policies` field, but the policies are actually "static", i.e. they are constant over the whole batch of agents, and `policies` array itself does not have a batch_dim. The consequence of this is that, given a batched Agent, one cannot e.g. `scan` or `vmap` over the agent object itself. Also, it's handy to be able to "slice out" an Agent from the batch e.g. doing `jtu.tree_map(lambda x: x[i:i+1, ...], agent)`. However, this is currently impossible without customly handling the `policies` field and its shape.

Potential fixes would be:
- mark `policies` as a static field, but Equinox will warn you that this is not recommended.
- broadcast `policies` to having a batch_dim, but then you'll waste memory for the same constant value.
- mark `policies` as a static field, and make it a non jnp.array type, i.e. a plain list of lists, that we can cast to `jnp.asarray` when we need it.

I think the last option is cleanest and least overhead?

Wdyt?

conorheins:
My thoughts are:

1. Regarding the inability to do this : `jtu.tree_map(lambda x: x[i:i+1, ...], agent)` -- are there work-arounds using the `is_leaf: Callable` argument to `jtu.tree_map`, that would allow you to still slice out Agents from the batch while handling the policies as static with respect to the call? So I'm imagining something that is, roughly/semantically: 

```python
def dont_count_policies(leaf):
     """ some function that returns True for every array-valued leaf that is not a policy """
    if is_policy_array(leaf):
        return False
    if is_normal_array_leaf(leaf):
        return True
sliced_agents = jtu.tree_map(lambda x: x[:i:i+1, ...], agent, is_leaf=dont_count_policies)
``` 

There may be other utilities in equinox worth looking into (into the tree filtering utilities) that accomplish the same thing or maybe even more cleanly/easily. Side thought: a possible way to define the `is_policy_array` function (which I left undefined above intentionally) would be to wrap the `policies` into a custom dataclass/equinox type like `PolicyMatrix(eqx.Module)`, so that that function could be a check on the type (`isinstance(leaf, PolicyMatrix)`) rather than some error-prone, custom logic like `leaf.shape == (planning_horizon, num_controls)`.

2. "mark policies as a static field, and make it a non jnp.array type, i.e. a plain list of lists, that we can cast to jnp.asarray when we need it." Seems like a good solution as well -- the only things to check would be 
[...]

PR #416, opened 2026-07-05
fix(agent): back Policies with a hashable tuple

### LINKED ISSUE #346: Static JAX arrays in Equinox static=True fields can break `eqx.filter_jit` / cause recompilation errors
### Problem
PR #345 puts a JAX array (formerly, what was called `policies` in the agent class) into an `equinox` static=True field. This can cause failures with `equinox.filter_jit` or other transforms that treat non-array parts as static arguments. Upon re-jit/reuse, Equinox/JAX sometimes need to compare/hash statics; JAX arrays don’t support that (array truth value is ambiguous; arrays aren’t hashable), which can crash or trigger pathological recompilation.

### Why we don't currently see it:
 Our tests/notebooks largely use jax.jit directly, or trace once; we don’t repeatedly call `filter_jit`’d functions with changing static fields.

### Future risks.

Code like “environment loop wrapped in a single filter_jit’d function” (an example that @dimarkov raised to me) or any pattern that reuses a `filter_jit`’d function with an Agent argument will eventually trigger this if any static field contains a JAX array.

```python
import equinox as eqx
import jax.numpy as jnp

from pymdp.agent import Agent


def build_agent():
    A = [jnp.array([[0.9, 0.1], [0.1, 0.9]], dtype=jnp.float32)]
    B = [jnp.stack([jnp.eye(2, dtype=jnp.float32), jnp.flipud(jnp.eye(2, dtype=jnp.float32))], axis=-1)]
    return Agent(A=A, B=B, policy_len=1, batch_size=1)


@eqx.filter_jit
def uses_agent(agent, x):
    # Access a value from the static policies field.
    return x + agent.policies.policy_arr[0, 0, 0]


def main():
    agent = build_agent()
    print("first", uses_agent(agent, 0.0))
    print("second", uses_agent(agent, 1.0))

    agent2 = build_agent()
    print("third", uses_agent(agent2, 2.0))


if __name__ == "__main__":
    main()
```

- With `JAX_LOG_COMPILES=1`, uses_agent recompiles on every call (even with the same agent), which indicates cache misses from static JAX arrays.
- A direct jax.jit(..., static_argnums=0) with a JAX array fails with `ValueError Non-hashable static arguments...`, matching the failure mode described above

gauraaansh:
Fixes #346

`Agent.policies` is marked `static=True` (#345) to support `vmap`/`scan` over a batched Agent (#227), since `policies` has no batch dimension. But `Policies.policy_arr` is a real `jnp.ndarray`, so the static partition ends up identity-based instead of value-based, and every fresh `Agent` construction is a cache miss under `eqx.filter_jit`.

This backs `Policies` with a hashable tuple instead, exposing `policy_arr` as a property that materializes a JAX array on access. `Agent.policies` stays static, but the static partition is now hashable by value.

Constructing an `Agent` inside an active `jax.jit` trace broke the first version of this, since a traced array can't be converted to a concrete tuple mid-trace. Traced this to `construct_policies()` converting to JAX arrays earlier than necessary. Added Python tuple builders that defer the JAX conversion to the property/return boundary.

## Testing

- `hash(Policies)` now works (`main` raises `TypeError: unhashable type 'ArrayImpl'`)
- `jax.jit(f, static_argnums=0)` with a `Policies` argument now works (`main` raises `ValueError: Non-hashable static arguments...`)
- The "A JAX array is being set as static!" warning no longer appears
- `vmap`/`scan`/slicing over a batched Agent still work (#227)
- New construction path verified equivalent to the old output.
- No memory regression observed under repeated Agent construction (CPU and GPU) or in a compiled workload mirroring the planning loop's access pattern (CPU). At larger policy-table sizes, the tuple implementation showed lower peak memory usage than the current implementation during repeated construction.
- Full test suite: 305 passed, 1 skipped, 1 pre-existing unrelated error

### DIFF
--- a/pymdp/agent.py
+++ b/pymdp/agent.py
@@ -211,14 +211,14 @@ class Agent(Module):
         if (
             policies is None and B_action_dependencies is not None
         ):  # note, this only works when B_action_dependencies is not the trivial case of [[0], [1], ...., [num_factors-1]]
-            policies_multi = control.construct_policies(
+            policies_multi_tup = control._construct_policies_tuple(
                 self.num_controls_multi,
                 self.num_controls_multi,
                 policy_len,
                 control_fac_idx,
             )
             B, pB, self.action_maps = self._flatten_B_action_dims(B, pB, self.B_action_dependencies)
-            policies = self._construct_flattend_policies(policies_multi, self.action_maps)
+            policies = self._construct_flattend_policies_tuple(policies_multi_tup, self.action_maps)
             self.sampling_mode = "full"
         
         # extract shapes from A and B
@@ -307,18 +307,23 @@ class Agent(Module):
 
         # construct policies
         if policies is None:
-            policies_array = control.construct_policies(
+            policies_tup = control._construct_policies_tuple(
                 self.num_states,
                 self.num_controls,
                 self.policy_len,
                 self.control_fac_idx,
             )
-            self.policies = control.Policies(policies_array)
+            self.policies = control.Policies(policies_tup)
         else:
-            if not isinstance(policies, control.Policies):
-                self.policies = control.Policies(jnp.array(policies))
-            else:
+            if isinstance(policies, control.Policies):
                 self.policies = policies
+            elif isinstance(policies, tuple):
+                # already a concrete nested tuple (e.g. from `_construct_flattend_policies_tuple`);
+                # pass it straight through so `Policies.__init__` doesn't need to round-trip it
+                # through `jnp.array` first, which would defeat the point under an active jit trace
+                self.policies = control.Policies(policies)
+            else:
+                self.policies = control.Policies(jnp.array(policies))
 
         if C is None:
             C = [jnp.ones((self.batch_size, self.num_obs[m])) / self.num_obs[m] for m in range(self.num_modalities)]
@@ -1135,6 +1140,47 @@ class Agent(Module):
             )
         policies_flat = jnp.stack(policies_flat, axis=-1)
         return policies_flat
+
+    def _construct_flattend_policies_tuple(
+        self, policies_tup: tuple, action_maps: list[dict[str, Any]]
+    ) -> tuple:
+        """
+        Pure-Python equivalent of `_construct_flattend_policies`, operating on and
+        returning nested tuples instead of `jnp.ndarray`. Mirrors
+        `utils.get_combination_index`'s mixed-radix encoding exactly, but never invokes
+        a JAX primitive, so it's safe to call while `Agent.__init__` is itself being
+        traced under `jax.jit` (see `control._construct_policies_tuple`'s docstring for
+        why that matters).
+        """
+        num_policies = len(policies_tup)
+        horizon = len(policies_tup[0])
+
+        columns = []
+        for action_map in action_maps:
+            multi_dependency = action_map["multi_dependency"]
+            if multi_dependency == []:
+                columns.append(tuple(tuple(0 for _ in range(horizon)) for _ in range(num_policies)))
+                continue
+
+            dims = action_map["multi_dims"]
+            column = []
+            for policy in policies_tup:
+                row = []
+                for t in range(horizon):
+                    xs = [policy[t][d] for d in multi_dependency]
+                    index = 0
+                    product = 1
+                    for i in reversed(range(len(dims))):
+                        index += xs[i] * product
+                        product *= dims[i]
+                    row.append(index)
+                column.append(tuple(row))
+            columns.append(tuple(column))
+
+        return tuple(
+            tuple(tuple(columns[a][pol][t] for a in range(len(action_maps))) for t in range(horizon))
+            for pol in range(num_policies)
+        )
 
     def _get_default_params(self) -> dict[str, Any] | None:
         method = self.inference_algo
--- a/pymdp/control.py
+++ b/pymdp/control.py
@@ -5,8 +5,10 @@
 """Policy construction, expected free energy, and action sampling utilities."""
 
 import itertools
+import jax
 import jax.numpy as jnp
 import jax.tree_util as jtu
+from collections import OrderedDict
 from typing import Sequence
 from functools import partial
 from jax import lax, vmap, nn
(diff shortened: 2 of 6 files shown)


conorheins:
Merged #417 -- @gauraaansh can you please rebase this branch to main and then we can retrigger CI?

pymdp/control.py
```diff
@@ -18,23 +18,46 @@
 from pymdp.maths import factor_dot, log_stable, stable_entropy, stable_xlogx, spm_wnorm
 from pymdp import utils
 
+def _to_nested_tuple(x):
```
conorheins: maybe better to put this in pymdp/utils.py?

pymdp/control.py
```diff
@@ -18,23 +18,46 @@
...
+    def __init__(self, policy_arr: Array | tuple) -> None:
+        if isinstance(policy_arr, tuple):
+            # Already a concrete nested tuple (e.g. from `_construct_policies_tuple`).
+            # Building it never invoked a JAX primitive, so it's safe to use directly
+            # even if this constructor is itself being called while `Agent.__init__` is
+            # traced under `jax.jit` -- there's no array here to concretize.
+            self._policy_tup = policy_arr
+            self.num_policies = len(policy_arr)
+            self.horizon = len(policy_arr[0])
+            self._dtype = jnp.int32
```
conorheins: with this enforcement of jnp.int32, x64 users will silently get int32 instead of int64, maybe worth fixing this to the dtype based on the jax dtype config to make it more generic?

pymdp/control.py
```diff
@@ -18,23 +18,46 @@
...
+        else:
+            self.num_policies = policy_arr.shape[0]
+            self.horizon = policy_arr.shape[1]
+            self._dtype = policy_arr.dtype
+            self._policy_tup = _to_nested_tuple(policy_arr.tolist())
+
+    @property
+    def policy_arr(self) -> Array:
+        return jnp.array(self._policy_tup, dtype=self._dtype)
```
conorheins: this will rebuild the whole table on every access, which according to my measurements will add about 10 ms of overhead at large policy counts (specifically tested 4096 policies) -- have you benchmarked this at large policy counts and what overhead this adds in the context of e.g. planning where this access will get called a lot? that would be helpful to know before we pull the trigger on this, when it could add non-trivial overhead for users doing long planning depths and many policies

pymdp/agent.py
```diff
@@ -1136,6 +1141,47 @@ def _construct_flattend_policies(self, policies: Array, action_maps: list[dict[s
         policies_flat = jnp.stack(policies_flat, axis=-1)
         return policies_flat
 
+    def _construct_flattend_policies_tuple(
```
conorheins, line 1144: this is nice, but could you also repurpose the old builder (`_construct_flattend_policies()`) as a test reference to create a unit test for this new function's correcntess?

conorheins (changes requested):
Thanks for this, this is great! I appreciate the amount of manual testing that clearly went into it. A few things I'd like addressed before merging, plus a couple of smaller notes (in addition to the inline comments):

## Please fix before merging

1. I wasn't able to independently replicate this "recompiles drop 3→1" claim in the PR description. I ran a controlled comparison against `main`, holding the jit'd function's other argument fixed as a `jnp.array` (rather than a Python scalar): `eqx.filter_jit` already gives 1 trace on `main` across fresh, value-identical `Agent`s. `eqx.Module.__eq__`/treedef equality is value-based, so two fresh agents with identical static arrays already compared equal pre-PR; there was no recompile-per-fresh-agent to fix. The fix is still worthwhile of course, just for different reasons than the reason of reducing recompiles. The benefits are of course that `hash(Policies)` now works (`main` raises `TypeError: unhashable type 'ArrayImpl'`), `jax.jit(f, static_argnums=0)` with a `Policies` argument now works (`main`: `ValueError: Non-hashable static arguments...`), and the per-construction "A JAX array is being set as static!" warning no longer appears. So given that, could you update the description to lead with hashability/robustness rather than the recompile-count number, assuming you agree with this assessment? Happy to be corrected if I'm not underestanding something though.

2. It would be good to add test coverage for the rewritten policy functions (`control._construct_policies_tuple` and `Agent._construct_flattend_policies_tuple`) because right now the existing suite doesn't flag if those functions are buggy.  I mutation-tested to check:
   - Patched `_construct_flattend_policies_tuple` to use forward instead of reversed mixed-radix order → for the exact config `test_agent_jax.py` already uses (`num_controls=[2,3,2]`, `B_action_dependencies=[[],[0,1],[0,2]]`), **10 of 12 policies change** — all 14 tests in that file still pass.
   - Patched `_construct_policies_tuple` with a non-bijective clamp that corrupts every default policy table and makes the highest action unreachable → **22/22 tests** across `test_agent_jax.py`, `test_control_jax.py`, `test_cue_chaining_env.py` still pass.

Existing assertions only check shapes, `q_pi` normalization, and finiteness but not policy-table content. Given the inline comment asking to keep the old builder around as a test reference, I'd suggest pairing that with an actual equivalence test (`jnp.array_equal` against the old jnp-based builder for both `construct_policies` and `_construct_flattend_policies`, over a couple of `policy_len > 1` / explicit `control_fac_idx` configs).

Related: no test asserts the bug this PR fixes is actually fixed; e.g. `hash(agent1.policies) == hash(agent2.policies)` for two identically-constructed agents fails on `main` today with `TypeError`, so it'd be a genuine regression test. Also you can remove the global filter of the "A JAX array is being set as static!" warning from `test/conftest.py`, which is exactly one of the symptoms of this bug.

3. Small hash/eq contract bug in `Policies.__init__`. The tuple-construction branch stores `self._dtype = jnp.int32` (a class), while the array-construction branch stores `policy_arr.dtype` (an `np.dtype` instance). These compare `__eq__`-equal but hash *unequal*:
   ```python
   p_tup = Policies(tup)                              # _dtype = jnp.int32
   p_arr = Policies(jnp.array(tup, dtype=jnp.int32))   # _dtype = dtype('int32')
   p_tup == p_arr             # True
   hash(p_tup) == hash(p_arr) # False
   {p_tup: 1}.get(p_arr)      # None -- silent miss
   ```
Doesn't break `jax.jit`'s cache (matching goes through `__eq__`/treedef equality, not raw hashing), but it's exactly the "equal objects, different cache identity" bug class this PR exists to eliminate — any hash-keyed cache would silently double-cache. One-line fix: normalize to `jnp.dtype(jnp.int32)` in the tuple branch so both paths produce the same type.

pymdp/control.py
gauraaansh: Yes this is introducing overhead exactly like you said, tested it at 4096 policies: 12.8ms same as you,
 
Should have cached this, Tried with caching keyed on the policy tuple and dtype, cached only concrete results, not a tracer so it's safe if this gets accessed while something is being traced under jit or scan. 

 num_policies │ uncached                            │ cached 

 4096         │ ~7.1ms                                         │ ~102us                           

 10,000       │ ~16.1ms                                        │ ~257us                           

 65,536       │ ~171.6ms                                       │ ~2.66ms                          
 
at 65,536 policies, a single cached access costs 2.66ms. 

An acceptable tradeoff at the scale ?  or should I try to push it further

gauraaansh:
You are completely right, Reran a comparison holding the jit'd function's other argument fixed as a jnp.array : 1 trace on main and on this branch, Theres no recompile-per-fresh-agent to fix. 

The original number came from the reproduction where that other argument was a raw Python float instead, which caused retraces unrelated to Policies entirely. Should have controlled for it before writing the claim in PR. Will update the description to lead with hashability instead: 

`hash(Policies)` works,
`jax.jit(f, static_argnums=0)` with a Policies argument works,
the static-array warning is gone. 

Adding equivalence tests for both tuple builders. `control._construct_policies_tuple` against an independent array-based reference, since the old jnp-based `construct_policies` doesn't exist as a separate function anymore.

`Agent._construct_flattend_policies_tuple` against the old `_construct_flattend_policies`, both through a full Agent build across configs and as a direct standalone call.

For the standalone call, `action_maps` are derived from the `_flatten_B_action_dims` , that  only reads `self.num_controls_multi`, so a lightweight stand-in object plus dummy-shaped B tensors gets real `action_maps` without a full Agent build. Kept a integration-style test alongside it too, since the isolated test only covers the tuple builder given correct input.

Adding the regression test you described: hash(agent1.policies) == hash(agent2.policies) for two identically constructed agents. Fails with TypeError on main, passes here. Removed the conftest.py warning filter and confirmed with -W error::UserWarning across the full suite that it's gone.

conorheins (approved):
Nice, this looks good to me. I'm approving, because the failing CI is due to this recurring XLA compile-hang issue which reappeared after release of jax 0.11.0 (see my draft PR #420 which should address it).

Only small comment would be cleaning up some of the long docstrings in the new functions added during review, which make references to my in-lien PR review comments -- that's a bit over-verbose and will load docstrings up with references to old PR review comments that will same out of place to future users/coding agents who haven't followed the lineage of this PR review. 

Not blocking for merge, but would be nicer to make some of those docstrings/comments a bit more concise, since they'll appear pretty outdated/out-of-place in the future

conorheins:
Thanks @gauraaansh; this is a big help and important contribution to pymdp :)

gauraaansh:
> Thanks @gauraaansh; this is a big help and important contribution to pymdp :)

most welcome, learning a lot from here.

PR #425, opened 2026-08-01
fix(legacy): make deterministic action/policy tie-breaks reproducible

docxology:
## Summary

The legacy API's `action_selection="deterministic"` path is not actually deterministic on ties: `select_highest` (`pymdp/legacy/control.py`) breaks exact (within 1e-8) ties via the module-global NumPy RNG (`np.random.choice`), so a seeded run is not reproducible and the call is not thread-safe.

This PR threads an optional `seed` parameter through the production `select_highest`, `sample_action` and `sample_policy` functions, using a local `np.random.default_rng(seed)` for the tie-break (the same pattern the existing `_select_highest_test` helper already used). Supplying a seed makes the result reproducible; `seed=None` keeps current behavior for backward compatibility.

It also corrects the `select_highest` docstring, which claimed a 1e-5 tolerance while the code uses 1e-8.

Finally, the two deterministic tie-break tests were rewritten to exercise the **real production paths** (`sample_action`/`sample_policy` with `seed=`) instead of the test-only `_sample_action_test`/`_sample_policy_test` helpers — those helpers were dead code never referenced by production, so the nondeterministic behavior had zero real coverage.

## Verification

- Legacy import with only the required deps confirmed that `opt_einsum` is a required (currently undeclared) import — separate concern, not part of this PR.
- `select_highest` tie behavior validated in isolation: same seed → same result; different seeds → subset of the tied indices; non-tie → argmax.
- Production `sample_action`/`sample_policy` with `seed=` reproduce exactly the values the old tests asserted (action: seed 1923 → 0, 48323 → 1; policy: seed 1923 → 1, 48323 → 2).
- Full `test/test_control.py` passes: **23 passed**.

## Tests

- `test/test_control.py::TestControl::test_deterministic_action_sampling_equal_value`
- `test/test_control.py::TestControl::test_deterministic_policy_selection_equal_value`

Both now call the production functions with an explicit seed.

### DIFF
--- a/pymdp/legacy/control.py
+++ b/pymdp/legacy/control.py
@@ -1014,7 +1014,7 @@ def get_num_controls_from_policies(policies):
     return list(np.max(np.vstack(policies), axis = 0) + 1)
     
 
-def sample_action(q_pi, policies, num_controls, action_selection="deterministic", alpha = 16.0):
+def sample_action(q_pi, policies, num_controls, action_selection="deterministic", alpha = 16.0, seed=None):
     """
     Computes the marginal posterior over actions and then samples an action from it, one action per control factor.
 
@@ -1057,7 +1057,7 @@ def sample_action(q_pi, policies, num_controls, action_selection="deterministic"
 
         # Either you do this:
         if action_selection == 'deterministic':
-            selected_policy[factor_i] = select_highest(action_marginals[factor_i])
+            selected_policy[factor_i] = select_highest(action_marginals[factor_i], seed=seed)
         elif action_selection == 'stochastic':
             log_marginal_f = spm_log_single(action_marginals[factor_i])
             p_actions = softmax(log_marginal_f * alpha)
@@ -1123,7 +1123,7 @@ def _sample_action_test(q_pi, policies, num_controls, action_selection="determin
 
     return selected_policy, p_actions
 
-def sample_policy(q_pi, policies, num_controls, action_selection="deterministic", alpha = 16.0):
+def sample_policy(q_pi, policies, num_controls, action_selection="deterministic", alpha = 16.0, seed=None):
     """
     Samples a policy from the posterior over policies, taking the action (per control factor) entailed by the first timestep of the selected policy.
 
@@ -1153,7 +1153,7 @@ def sample_policy(q_pi, policies, num_controls, action_selection="deterministic"
     num_factors = len(num_controls)
 
     if action_selection == "deterministic":
-        policy_idx = select_highest(q_pi)
+        policy_idx = select_highest(q_pi, seed=seed)
     elif action_selection == "stochastic":
         log_qpi = spm_log_single(q_pi)
         p_policies = softmax(log_qpi * alpha)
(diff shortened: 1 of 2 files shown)


pymdp/legacy/control.py
```diff
@@ -1229,7 +1231,8 @@ def select_highest(options_array):
                     abs(options_with_idx[:, 1] - np.amax(options_with_idx[:, 1])) <= 1e-8][:, 0]
     if len(same_prob) > 1:
         # If some of the most likely actions have nearly equal probability, sample from this subset of actions, instead of using argmax
-        return int(same_prob[np.random.choice(len(same_prob))])
+        rng = np.random.default_rng(seed)
```
Arun-Niranjan, line 1234: @conorheins I think I'm right in saying that we should avoid using numpy rng, and instead be using JAX rng instead?

If so, I'll create an issue for updating the CONTRIBUTING.md file to include this.

conorheins:
Will review this soon,  related to an older PR made by @AleMuzzi (iirc) a couple years ago where this deterministic breaking due to matched probs was introduced,  will reference thay during review

pymdp/legacy/control.py
conorheins: Well yes in general but this is addressing legacy, where numpy rng was used.  Bigger thing that needs dealing with is just how we handle PRs thay touch legacy code in general. We should decide whether we will keep numpy legacy backend up to date, and if we do decide yes, we should detail thay decision in CONTRIBUTING.md as you say

conorheins (changes requested):
Thanks @docxology. The docstring fix and moving the tests onto the production functions are both good, but the seeding mechanism itself needs rework before this can go in.

The problem: `select_highest` builds `np.random.default_rng(seed)` fresh on every call, and a newborn generator's first draw is always the same. So a fixed seed doesn't make the tie-break reproducible, it makes it *constant*:

```python
q = np.ones(3) / 3   # 3-way tie
Counter(int(sample_action(q, policies, [3], seed=42)[0]) for _ in range(100))
# Counter({0: 100})
```

Same index at every timestep. For context on why that matters: the whole reason this tie-break exists is #115 / #116 / #117 -- deterministic argmax was systematically picking index 0 among equi-probable actions, and @AleMuzzi's fix made it sample among the tied subset instead. Pre-PR behavior under `np.random.seed(42)` gives `{0: 33, 1: 36, 2: 31}` over 100 calls -- reproducible *and* varying. This PR brings the frozen-index pathology back for any seeded run, just at a seed-dependent index instead of 0. It also correlates across control factors (every factor with the same tie multiplicity picks the same index, since each gets the same fresh rng).

Two more issues:

1. `seed=None` doesn't preserve current behavior like the docstring says. `default_rng(None)` pulls OS entropy and ignores `np.random.seed`, so every existing user who seeds globally at the top of a script -- which is the only reproducibility mechanism the legacy Agent path has, since `Agent.sample_action()` can't pass a seed -- silently loses it.

2. The seed is silently ignored on the `action_selection="stochastic"` branches, which still go through the global RNG via `utils.sample`.

The fix shape I'd like: accept `rng: np.random.Generator | int | None` instead of an int seed. `None` -> keep using the global `np.random` exactly as today (real backward compat); a Generator -> use it, so state advances across calls. Then thread it through the Agent: accept `seed` in the constructor, store `self.rng = np.random.default_rng(seed)` **once**, and pass it down in `sample_action`. That gives reproducible-across-runs *and* varying-within-run, and makes the feature reachable from the API people actually use. Funnily enough your current code already supports this by accident -- `default_rng` passes a Generator through unchanged -- it's just undocumented and the int path is the footgun.

On tests: one call per seed asserting a magic index can't catch any of this (a seed-ignoring implementation passes each assertion as a coin flip). Better: same seed, rebuild the Generator, assert the full sequence matches; plus something like 50 calls with one Generator hitting >=2 distinct tied indices -- deterministic given the seed, so not flaky, and it pins the actual #115 property.

@Arun-Niranjan re: our RNG thread -- jax rng shouldn't go anywhere near `pymdp/legacy/`, that backend exists so people can run without jax at all. The convention I want in CONTRIBUTING.md: legacy = numpy Generator idioms (what this PR should be doing), current API = PRNGKey threading, and legacy is maintenance-mode -- bugfixes like this one welcome, new features go to the jax side. ill write that up later

PR #442, opened 2026-08-08
fix: label lists on multiple axes select blocks instead of diagonals

### LINKED ISSUE #424: Distribution.set writes the diagonal when given a list of labels on two axes
Filling a tensor through the labelled API can write the wrong cells without
saying so. `Distribution.set` given a list of labels on two axes writes the
diagonal of the block those labels describe, and `normalize()` will happily
renormalise the result, so a wrong generative model can reach an agent with
nothing raising anywhere along the way.

The cause is that `get` and `set` hand their index tuple to numpy and inherit
its advanced-indexing rules, which the labels do not imply. There are two
separate effects. I could not find anything that says which behaviour is
intended, so part of this is a question.

Operating system: Ubuntu 24.04, Linux 6.17.0. Python 3.12.3, numpy 2.5.1,
pymdp main at 0681ce4, installed with `pip install -e .` into a clean venv.

## 1. A list on two axes selects the diagonal, not the block

```python
import numpy as np
from pymdp.distribution import Distribution

B = Distribution({"obs": ["red", "green", "blue"]},
                 {"state": ["s0", "s1", "s2"]},
                 data=np.zeros((3, 3)))

B.set(event={"obs": ["red", "green"]}, batch={"state": ["s0", "s1"]}, values=7.0)
print(B.data)
# [[7. 0. 0.]
#  [0. 7. 0.]
#  [0. 0. 0.]]
```

Two cells written where four were named, and no error. Writing an array instead
of a scalar does raise, but about something else:

```python
B.set(event={"obs": ["red", "green"]}, batch={"state": ["s0", "s1"]},
      values=np.array([[1.0, 2.0], [3.0, 4.0]]))
# ValueError: shape mismatch: value array of shape (2,2) could not be
# broadcast to indexing result of shape (2,)
```

Reading has the same shape of problem:

```python
A = Distribution({"obs": ["red", "green", "blue"]},
                 {"state": ["s0", "s1", "s2"]},
                 data=np.arange(9.0).reshape(3, 3))

print(A.get(event={"obs": ["red", "green"]}, batch={"state": ["s0", "s1"]}))
# [0. 4.]        shape (2,)

print(A.data[np.ix_([0, 1], [0, 1])])
# [[0. 1.]
#  [3. 4.]]      shape (2, 2)
```

## 2. An axis in between reorders the result

This one needs only a single list:

```python
D = Distribution({"obs": ["o0", "o1", "o2"]},
                 {"state": ["s0", "s1", "s2", "s3"], "ctrl": ["up", "down"]},
                 data=np.arange(24.0).reshape(3, 4, 2))

print(D.get(event={"obs": "o0"}, batch={"ctrl": ["up", "down"]}).shape)
# (2, 4)         declared order is obs, state, ctrl, so this should be (4, 2)
```

A scalar index counts as an advanced index for this rule, so `o0` and the
`ctrl` list are two advanced indices separated by the `state` slice, and numpy
moves the broadcast dimensions to the front. The labels say nothing about that.

## What I expected

Labels on two axes describe a sub-block, and the axes should come back in the
order they were declared. That is what the labelled-array libraries this API
resembles do, which I checked rather than assumed:

```python
import xarray as xr
da = xr.DataArray(np.arange(9.0).reshape(3, 3),
                  coords={"obs": ["red", "green", "blue"],
                          "state": ["s0", "s1", "s2"]},
                  dims=["obs", "state"])
da.sel(obs=["red", "green"], state=["s0", "s1"])     # (2, 2), the block

import pandas as pd
df = pd.DataFrame(np.arange(9.0).reshape(3, 3),
                  index=["red", "green", "blue"], columns=["s0", "s1", "s2"])
df.loc[["red", "green"], ["s0", "s1"]]               # (2, 2), the block
```

On the three-axis case in section 2, xarray gives `(4, 2)` with dims
`('state', 'ctrl')`, which is the declared order with the scalar axis dropped.
Two lists there, `sel(obs=["o0","o1"], ctrl=["up","down"])`, give `(2, 4, 2)`
with dims `('obs', 'state', 'ctrl')`. Neither gets reordered.

## Cause

`get` and `set` concatenate the per-axis results of `_get_slices` and pass the
tuple straight through:

```python
slices = event_slices + batch_slices
return self._data[tuple(slices)]
```

`_get_slices` turns labels into integers and slices, and from there both
effects are ordinary numpy advanced indexing. Two index arrays broadcast
against each other rather than forming an outer product, and a slice between
advanced indices moves the broadcast dimensions to the front. Neither is
documented; `get` and `set` have no docstrings.

## Why part of this is a question

`__getitem__` and `__setitem__` reject lists outright, on any axis:

```python
A[["red", "green"], ["s0", "s1"]]
# TypeError: unhashable type: 'list'
```

So the two indexing paths disagree about whether multi-label selection is
supported at all, and the tutorial the docs link to fills tensors through
`__setitem__` only. Nothing in `pymdp/`, `test/`, `examples/` or `docs-mkdocs/`
passes a list to `get` or `set`, and the branch in `_get_slices` that handles
one has no test coverage. It has been in the file since ad9c098.

There is a coherent reading where none of this is a defect: labels are a naming
layer over numpy indexing, a list is a numpy fancy index, and the pointwise
gather is the feature. I do not think that is what someone filling a `B` tensor
by label expects, but you may have meant it, and #26 reads like the tension was
deliberate.

## If you do want it changed

I have a change for the first effect: reshape each list onto its own axis, the
way `np.ix_` does, leaving single lists, scalars and slices alone. `np.ix_`
cannot be called directly here because it rejects `slice(None)`.

I added four tests to `test_distribution.py`. Three of them fail on 0681ce4 and
all pass with the change. Running `pytest test/test_distribution.py
test/test_utils.py test/test_control.py test/test_wrappers.py` gives 29 passed
on 0681ce4 as it stands, and 33 passed with the four tests and the change. The
29 that exist today pass either way.

It does not fix the second effect. On the tutorial's own `B` tensor,
`get(event={"position": [...]}, batch={"movement": [...]})` still returns
`(2, 2, 4)` where the declared order gives `(2, 4, 2)`, and `set` correspondingly
demands the transposed value array. Fixing that means materialising the
intervening slices as index arrays or moving the axes back afterwards, which is
a bigger change than I wanted to make before asking.

So, two options as I see them, and I am happy to write either. Make label lists
select orthogonally and preserve declared order, matching xarray, or reject a
list on more than one axis with an error that points at `.data` and `np.ix_`
and document that labels index one element per axis. The second is a behaviour
change too, since the code returns an array today, but a much smaller one.

Arthur031221:
Fixes #424.

Implements the conversion described there: when any axis receives a list,
every axis becomes an index array. Slices turn into arange, each list is
reshaped onto its own output dimension, and no slices are left in the tuple,
so there is nothing for numpy to separate. Two lists now select the block
they name, and the declared axis order survives an untouched axis sitting
between advanced indices, which also fixes the reordering effect from the
issue. The gate is "at least one list present", so plain label and slice
access stays basic indexing, and a test pins that it still returns views.

The zip behaviour stays available where it is the point rather than the bug.
get and set take pointwise=True, and a points accessor covers bracket style
access:

```python
B["s1"].points[["A", "B"], ["C", "D"], "up"] = [0.9, 0.8]
```

Lists zip along one shared axis placed where the first list appears, the
remaining axes keep declared order, and ragged lists raise ValueError instead
of broadcasting into something surprising. The banded transition from the
issue thread works and is in the tests:

```python
transition.set(
    event={"loc": ["l0", "l0", "l1", "l2"]},
    batch={"loc": ["l0", "l1", "l2", "l3"], "ctrl": "up"},
    values=1.0, pointwise=True,
)
```

Bracket indexing takes lists as well, with the same block default, so
`d[["r", "g"], ["s0", "s1"]]` and the equivalent get call select the same
entries. Both bracket paths share one positional resolution step, so `d[...]`
and `d.points[...]` resolve the same keys against the same axes and reject the
same arities. They still differ in what they require: points needs a list on
at least one axis and raises without one, while brackets take a selection of
labels and slices alone. pointwise is keyword only on get and set, because
values is the third positional argument of set and pointwise was the third of
get, so mirroring a get call in set bound True to values. Indexing more axes
than the distribution has raises one named error from both bracket paths,
where the message used to depend on the path and on the kind of extra index.

One migration hazard is documented on set and pinned by a test. With block
selection, `set(event={"loc": locs}, batch={"loc": locs, "ctrl": "up"},
values=np.ones(4))` broadcasts the value into every row and fills all 16
entries of the 4 by 4 block. The same expression wrote the 4 diagonal entries
before, and pointwise=True still writes them.

Nineteen tests added to TestDists: the two axis block with a set round trip,
declared order across an untouched axis for get and for set, the view
guarantee on the no list path for get and for brackets, pointwise get and
set, the banded example, the points round trip, the zip axis position with a
slice in the selection, values orientation for a pointwise set whose first
list follows a slice, one list with pointwise=True, int and mixed labels
through points, an event and a batch axis sharing a key name with different
label orders, blocks through brackets with a set round trip and declared
order, the broadcast above against the diagonal pointwise=True writes,
pointwise refused positionally, and the ragged, no list and too many indices
error cases.

This branch's test file run against main's source gives 7 passed and 16
failed, against 23 passed here. ruff check with the repo config and the
version pinned in #441 (0.15.22 here) reports no findings on main or on this
branch.

### DIFF
--- a/pymdp/distribution.py
+++ b/pymdp/distribution.py
@@ -39,21 +39,120 @@ class Distribution:
                 raise ValueError(f"When setting the tensor directly, the new shape {value.shape} must match the existing tensor shape {self._data.shape}")
         self._data = value
 
-    def get(self, batch: dict | None = None, event: dict | None = None) -> np.ndarray:
+    def get(
+        self, batch: dict | None = None, event: dict | None = None, *, pointwise: bool = False
+    ) -> np.ndarray:
+        """Read the block the labels name, in declared axis order.
+
+        Lists on several axes select the block they describe. pointwise=True
+        zips them into single points instead, as described on set.
+        """
         event_slices = self._get_slices(event, self.event_indices, self.event)
         batch_slices = self._get_slices(batch, self.batch_indices, self.batch)
 
         slices = event_slices + batch_slices
-        return self._data[tuple(slices)]
+        return self._data[self._index(slices, pointwise)]
 
     def set(
-        self, batch: dict | None = None, event: dict | None = None, values: np.ndarray | float | int | None = None
+        self,
+        batch: dict | None = None,
+        event: dict | None = None,
+        values: np.ndarray | float | int | None = None,
+        *,
+        pointwise: bool = False,
     ) -> None:
+        """Write values into the block the labels name, in declared axis order.
+
+        values is broadcast into that block by the usual numpy rules, so a
+        value whose shape is smaller than the block is repeated across it.
+        Writing lists of labels on two axes therefore fills the whole block:
+
+            transition.set(
+                event={"loc": locs}, batch={"loc": locs, "ctrl": "up"},
+                values=np.ones(4),
+            )
+
+        writes 1.0 to all 16 entries of the 4 by 4 block, because a (4,)
+        value is broadcast into every row of it. Before block selection
+        existed this same expression wrote the 4 diagonal entries. Pass
+        pointwise=True to keep that meaning: the lists are then zipped into
+        4 points and values is matched against those points instead.
+        """
         event_slices = self._get_slices(event, self.event_indices, self.event)
         batch_slices = self._get_slices(batch, self.batch_indices, self.batch)
 
         slices = event_slices + batch_slices
-        self._data[tuple(slices)] = values
+        self._data[self._index(slices, pointwise)] = values
+
+    def _index(self, slices: list[Any], pointwise: bool) -> tuple:
+        """Turn per-axis indices into a tuple that selects what the labels name.
+
+        Without lists this returns the entries untouched, so plain label and
+        slice access stays basic indexing and keeps returning views.
+
+        With a list on any axis, every axis is converted to an index array:
+        slices become arange and each list is reshaped onto its own output
+        dimension. No slices are left in the tuple, so numpy has nothing to
+        separate and the result keeps the declared axis order. Two lists select
+        the block they describe rather than its diagonal. Scalars stay scalar
+        and their axes collapse, as with basic indexing.
+
+        With pointwise=True the lists zip instead: they must share one length,
+        they are consumed together along one output axis placed where the first
+        list appears, and the remaining axes keep declared order. This is the
+        old broadcast behaviour made explicit, for writing diagonals and
+        per-point updates that block selection cannot express.
+        """
+        has_list = any(isinstance(s, list) for s in slices)
+        if not has_list:
+            if pointwise:
+                raise ValueError("pointwise indexing needs a list of labels on at least one axis")
+            return tuple(slices)
+
+        if pointwise:
+            lengths = {len(s) for s in slices if isinstance(s, list)}
+            if len(lengths) > 1:
+                raise ValueError(
+                    f"pointwise lists must all have the same length, got lengths {sorted(lengths)}"
+                )
+            n_out = 1 + sum(1 for s in slices if isinstance(s, slice))
+        else:
+            n_out = sum(1 for s in slices if isinstance(s, (list, slice)))
+
+        out: list[Any] = []
+        dim = 0
+        point_dim = None
+        for axis, s in enumerate(slices):
+            if isinstance(s, list):
+                if pointwise:
+                    if point_dim is None:
+                        point_dim = dim
+                        dim += 1
+                    use = point_dim
+                else:
+                    use = dim
+                    dim += 1
+                shape = [1] * n_out
+                shape[use] = -1
+                out.append(np.asarray(s, dtype=int).reshape(shape))
+            elif isinstance(s, slice):
+                shape = [1] * n_out
+                shape[dim] = -1
+                out.append(np.arange(self._data.shape[axis]).reshape(shape))
+                dim += 1
+            else:
+                out.append(s)
+        return tuple(out)
+
+    @property
+    def points(self) -> "_PointwiseView":
+        """Positional pointwise indexing, the zip counterpart of block access.
+
+        d.points[["A", "B"], ["C", "D"], "up"] reads or writes the two entries
+        (A, C, up) and (B, D, up) rather than the 2 by 2 block. Lists zip along
+        one shared axis and must have equal lengths.
+        """
+        return _PointwiseView(self)
 
     def _get_slices(self, keys: dict | None, indices: dict, full_indices: dict) -> list[Any]:
         slices = []
@@ -77,7 +176,7 @@ class Distribution:
         else:
             return index_map[key]
 
-    def _get_index_from_axis(self, axis: int, element: Any) -> int | slice:
+    def _get_index_from_axis(self, axis: int, element: Any) -> int | slice | list[int]:
         if isinstance(element, slice):
             return slice(None)
         if axis < len(self.event):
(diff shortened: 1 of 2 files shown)


Arthur031221:
The red `lint` check is not from this branch. That run executed before #441 landed, so the job still ran `uvx ruff check` unpinned and picked up ruff 0.16, whose broadened default rule set reports 530 findings across the repository. With the range #441 pins, `ruff>=0.15,<0.16`, `ruff check` passes on this branch with no findings at all, measured on 0.15.22.

For completeness, under ruff 0.16.2 the two touched files report 26 findings on this branch and the same 26 on main. One of the 26, `PYI041` at `pymdp/distribution.py:55`, sits on a line this change adds: the `values: np.ndarray | float | int | None` annotation, where 0.16 prefers plain `float`. The pinned 0.15 series does not report it. Happy to narrow that annotation if you would rather the file be clean under the newer rules, but I left it alone rather than edit the branch for a rule the project does not run yet.

A plain re-run would not clear the check, since it would reuse a merge commit that predates the pin, so I have rebased onto current main (`c19fa9e`). Main has moved only by #441 since the base commit here, and that touched `lint.yaml`, `.pre-commit-config.yaml` and `pyproject.toml` only, so the code is unchanged by the rebase. `test/test_distribution.py` is 13 passed on the rebased branch. The run may need approval again.

One design point I would rather settle than leave implied. In your comment the two lines were:

```python
B["s1"][["A","B"], ["C","D"], "up"] = [[.9,.2],[.1,.8]]     # block, default
B["s1"].points[["A","B"], ["C","D"], "up"] = [.9, .8]       # specifies a hyperdiagonal, opt-in
```

The second line works on this branch. The first still raises `TypeError: unhashable type: 'list'`, because `Distribution.__getitem__` and `__setitem__` resolve every key through `_get_index_from_axis`, which hashes it. I left that alone because the default bracket semantics were the question you put to @riddhipits, @alec-tschantz and @tverbele, and I did not want the PR to answer it for them.

The effect is an asymmetry that is probably not what anyone wants long term: block selection is reachable only through `get` and `set`, while pointwise selection is reachable through both `get` and `set` and brackets. Making `__getitem__` and `__setitem__` accept lists is small, since `_get_index_from_axis` would collect a list the way `_get_slices` already does and the result would route through the same `_index` helper with `pointwise=False`. Roughly 10 lines plus tests.

Tell me which you prefer: add it to this PR now, leave it out and open a follow-up once the semantics question has answers, or leave brackets rejecting lists permanently and treat `get`, `set` and `.points` as the whole surface. If nobody has strong views I would add it here, since it makes the default consistent across both access styles.

conorheins:
I will review soon, thanks!

conorheins (approved):
real review below, this one accidentally i pushed 'accept changes' mean to click req changes

conorheins (changes requested):
Thanks @Arthur031221, this looks good. I pulled it down and verified your claims -- tests pass. Approving CI now.

On your question: go with adding it to this PR. Nobody's weighed in on the bracket semantics and I don't want to block a correct fix on silence -- brackets should accept lists with the same block default as `get`/`set`. The current asymmetry (zip reachable through brackets via `.points`, block not at all) is the same two-paths-disagree problem your issue opened with.

A few small changes while you're at it:

1. Make `pointwise` keyword-only (`*, pointwise=False`) in both `get` and `set`. It's currently 3rd positional in `get` but 4th in `set`, so someone who learns `d.get(batch, event, True)` and mirrors it with `d.set(batch, event, True)` binds `True` to `values` and silently writes 1.0 over the block. Nothing passes it positionally so this costs nothing.

2. One migration hazard worth documenting: the old diagonal idiom `set(event={"loc": locs}, batch={"loc": locs, ...}, values=np.ones(4))` now broadcasts `(4,)` across the `(4,4)` block -- full block of ones, no error. Every other wrong orientation raises; this is the only silent one. I don't want a guard (broadcasting into a block is normal numpy behavior), but please pin the broadcast semantics with a test and call it out in the docstring, pointing at `pointwise=True`.

3. The `_PointwiseView` docstring says "labels, integers, lists of either, and bare slices are accepted" but scalar-only access raises ValueError -- add the at-least-one-list qualifier. And a one-line arity check in `_translate` would turn the cryptic `IndexError: tuple index out of range` from an extra trailing slice into a clear too-many-indices error.

4. Extra tests: declared order for `set` (currently only pinned for `get`), values orientation for pointwise `set` when a slice precedes the first list, one-list `pointwise=True`, int/mixed labels through `.points`, and `assertRaisesRegex` on the two new ValueErrors (repo convention).

The typo'd-key silent write (`event={"obbs": ...}` writes the whole tensor) stays out of scope, I'll open a separate issue once this lands.

Arthur031221:
Thanks, that answers it. Brackets are in this PR now, along with all four
changes.

Brackets: `__getitem__` and `__setitem__` take lists with the block default,
so `d[["r", "g"], ["s0", "s1"]]` and the equivalent get call select the same
entries. Without a list the bracket path is still basic indexing and still
returns a view, which is pinned.

1. `pointwise` is keyword only on get and set. Nothing in the tree passed it
positionally, so the visible effect is that `d.get(batch, event, True)` is now
a TypeError, which means there is no working form to learn and mirror.

2. Broadcast pinned in `test_block_set_broadcasts_values_across_the_block`:
the `np.ones(4)` call fills all 16 entries of the block and leaves the other
control plane at zero, and the same call with `pointwise=True` writes the 4
diagonal entries. The set docstring carries that example, says what the
expression used to do, and points at `pointwise=True`. One correction to your
description: `(4,)` is not the only silent shape. `(4,1)` and `(1,4)` also
pass without an error and also fill the whole block, while `(2,)` and `(3,4)`
raise. So the silent set is every shape that broadcasts into the block, not
the length 4 vector alone.

3. The `_PointwiseView` docstring now carries the at-least-one-list qualifier
and sends scalar only access to `d[...]`. On the arity check I did not do
quite what you asked. Instead of a line inside `_translate`, the positional
resolution `_translate` was doing is now a method on Distribution,
`_positional_slices`, brackets use it too, and the check lives there.
`_translate` is a call to it. That is what makes `d[...]` and `d.points[...]`
accept the same keys and reject the same arities, which is the two paths
disagreeing again. Both now raise `IndexError: too many indices for
Distribution: it has 2 axes, but 3 were indexed`. IndexError rather than
ValueError, since that is what numpy raises for the same mistake on a plain
array. Worth being exact about the gain: the one case that was already clear
is the trailing slice through brackets, where numpy said `too many indices for
array`, so there the wording changed and the clarity did not. A trailing label
or int gave `list index out of range` out of the key lookup on both paths, and
those do get clearer.

4. Tests added: declared order for set, values orientation for a pointwise set
whose first list follows a slice (pinned through the `.points` form and
through `set(..., pointwise=True)`, including a two list selection where
zipping and block selection give different shapes, so the set path carries the
test rather than riding on `.points`), one list with
`pointwise=True`, int and mixed labels through `.points`, blocks and declared
order through brackets, the view guarantee through brackets, and pointwise
refused positionally. `test_pointwise_ragged_lists_raise` is now
`assertRaisesRegex` on both ValueErrors and covers scalar only access to
`.points`.

Measured on the commit being pushed:

- `test/test_distribution.py`: 23 passed on this branch. This branch's test
  file run against main's source gives 7 passed and 16 failed. Run against the
  commit you reviewed (a497d46) it gives 19 passed and 4 failed, the four
  being brackets with lists, brackets keeping declared order, pointwise
  refused positionally, and the axis count in the too many indices message.
- `pytest test -n 2 -m "not nightly"`: 323 passed, 78 subtests passed, plus
  one failure and one collection error that are a pybefit version mismatch
  in my environment (`pymdp_likelihood() got an unexpected keyword argument
  'task'`, and `default_dict_numpyro_svi` missing from `pybefit.inference`).
  Both reproduce unchanged on main, and neither touches distribution.py.
- ruff with the repo config and the pin from #441 (0.15.22 here): no findings
  on main or on this branch.

One thing you will run into if you probe the bracket API, unchanged by this
PR and present on main as well: a bounded slice is treated as a full one, so
`d[1:2]` gives what `d[:]` gives. `_get_index_from_axis` maps every slice to
`slice(None)`. I left it alone as out of scope here.

Two housekeeping notes. The commit message keeps every sentence you read and
adds one paragraph at the end covering the three behaviour changes. The
description has three changes against the version you read: the paragraph
saying bracket indexing rejects lists is replaced, a paragraph on the
broadcast hazard is added, and the closing paragraphs re-derive the test list
and the counts, which quoted ruff 0.16.2 from before #441 pinned it out. And
since this is a force push over the commit whose workflow run you approved,
the checks will probably be back to waiting for approval on the new sha.

conorheins (approved):
Looks good. All four changes landed correctly: keyword-only enforced, arity check gives the clean message for trailing slice/label/int, broadcast contract documented and pinned, and the shared `_positional_slices` helper makes both bracket paths use the same arity rejection.

One tiny test gap worth considering: `test_too_many_indices_names_the_axis_count` only exercises trailing *slice* indices (where slices short-circuit before label lookup), so a mutant that moves the arity check after resolution keeps all 23 tests green. A one-line assertion with a non-slice surplus index (e.g. `d["r", "s0", "s1"]`) pins it. Not a blocker — the other 28 mutants against the new logic all die cleanly — but worth a line if you're in there. Otherwise, ready to merge once CI is green.

Arthur031221:
Added, as `2dfc84d5b` on top of the commit you reviewed, so that commit is still in the history and your approval is still on it. The force push above it is me correcting a sentence in my own commit message that was wrong about how many assertions the test already had; same tree, and `c771502d5` was not touched.

I built your mutant first. Moving the `len(indices) > n_axes` check in `_positional_slices` below the resolution comprehension leaves `test/test_distribution.py` at 23 passed. Adding the assertion to that same mutated copy gives 1 failed, 22 passed:

```
AssertionError: "it has 2 axes, but 3 were indexed" does not match "list index out of range"
```

The frame that raises is `key = list(self.batch.keys())[axis - len(self.event)]`. It is the surplus axis index running off the list of batch key names, not the surplus label being looked up, so the label never gets resolved either way. The unmutated branch with the assertion is 23 passed.

The line is in `test_too_many_indices_names_the_axis_count`:

```python
with self.assertRaisesRegex(IndexError, "it has 2 axes, but 3 were indexed"):
    likelihood["r", "s0", "s1"]
```

Both bracket paths go through `_positional_slices`, so `.points` depends on the same ordering.

On the new commit: `pytest test/test_distribution.py` is 23 passed, and `pytest test/ -q -p no:randomly` is 324 passed with 78 subtests, plus 1 failure and 1 collection error that pristine main at 5848360 also has on the same command, a pybefit version mismatch in my environment. `ruff check` on 0.15.22 passes on both. The subtests figure prints under `-q` only.

The new sha has only the title check on it so far; the build and lint jobs have not started, so they are waiting on your approval to run.

PR #443, opened 2026-08-14
feat(sparse): Enable Sparse B (BCOO) in Agent constructor and FPI inference

docxology:
# feat: Enable Sparse B (BCOO) transition matrices in `Agent` with `fpi` inference

## Overview
`pymdp`'s JAX backend contains sparse contraction machinery in `maths.py` (`spm_dot_sparse`, `factor_dot_flex`), sparse belief handling in `inference.py` (`JAXSparse`), and sparse-dependency FPI in `algos.py`.

However, sparse `B` matrices (using JAX `sparse.BCOO`) previously failed at the `Agent` API boundary on 3 specific operations:
1. `pymdp.agent.Agent.__init__` (line 263): `jnp.broadcast_to` raised `TypeError` on unbatched BCOO tensors.
2. `pymdp.utils.validate_normalization` (line 109): `jnp.sum(tensor, axis=axis)` raised `TypeError: sum requires ndarray or scalar arguments, got BCOO`.
3. `pymdp.control.compute_expected_state` (line 324): contracted $B$ to a BCOO single-factor marginal belief vector, which subsequently broke downstream 1D belief propagation in `factor_dot`.

## Changes
- `pymdp/utils.py`: Added sparse summation branch in `validate_normalization` using `sparse.sparsify(jnp.sum)(tensor, axis=axis)`.
- `pymdp/control.py`: Wrapped the returned marginal from `compute_expected_state` with `maths._to_dense_if_sparse` to ensure 1D belief vectors remain dense arrays.
- `pymdp/agent.py`: Extended batch-dimension validation and error reporting for sparse arrays.

## Verification & Benchmarks
- Bit-identical accuracy against dense Agent: max posterior difference `0.000e+00`.
- Memory reduction: On a 38,200-state transition, resident memory during inference drops from 5.84 GB (dense) to 66.8 MB (sparse B), an **88x reduction**.

Issue #434, opened 2026-08-01
agent sampling_mode silently overwritten to 'full'


docxology:
## Summary

`pymdp/agent.py:222` silently overwrites the user's `sampling_mode` with `"full"` when `policies is None and B_action_dependencies is not None`, with no warning — so a caller who requested `sampling_mode="marginal"` gets `"full"` behavior without being told.

## Details

```python
# agent.py:222 (inside __init__)
if policies is None and B_action_dependencies is not None:
    ...
    self.sampling_mode = "full"
```

`sampling_mode` is a constructor parameter (default `"marginal"`, assigned at `agent.py:286`). When flattened `B_action_dependencies` force the full-policy sampling mode, the user's requested value is discarded silently — a classic silent-semantic-override trap for multi-action agents.

## Suggested fix

Compare against the requested value before overriding and emit a warning (or raise), e.g.:

```python
if policies is None and B_action_dependencies is not None:
    if self.sampling_mode != "full":
        warnings.warn(
            "Flattened B_action_dependencies require sampling_mode='full'; "
            f"overriding requested sampling_mode={self.sampling_mode!r}."
        )
        self.sampling_mode = "full"
```

## Environment

- inferactively-pymdp 1.0.3 (commit 0681ce4)
- Found via static red-team review

PR #445, opened 2026-08-16
fix: harden modern inference, learning, and dependency contracts

docxology:
## Summary

- Prevent masked inference from mutating `Agent.A` across calls.
- Validate unsupported modern Agent sampling/action configuration at construction.
- Fail loudly for currently unsupported modern learning flags.
- Define zero-mass normalization behavior instead of returning NaNs.
- Preserve structural-zero support during modern A/B Dirichlet learning.
- Resolve omitted multi-factor observation dependencies consistently.
- Declare the runtime `numpyro` dependency used by `pymdp.likelihoods`.
- Fix a vacuous Distribution assertion and add direct likelihood import coverage.
- Make the optional pybefit integration skip explicit and actionable.

## Verification

- `uv run --with pytest --with pytest-xdist python -m pytest -q`
  - 317 passed, 1 skipped, 78 subtests passed, 3 warnings
- `uv run ruff check pymdp test`
  - passed
- `python -m py_compile` on all touched Python files
  - passed
- `git diff --check`
  - passed

The full test run was performed in a clean submission worktree based on current `upstream/main` (`c19fa9e`) with the two commits in this branch.

## Scope notes

This PR intentionally does not duplicate the implementation work in open PR #443 (sparse B integration) or the existing dependency changes in #426; it adds the independent runtime dependency and correctness/test-contract fixes verified by this audit.

### DIFF
--- a/pymdp/agent.py
+++ b/pymdp/agent.py
@@ -176,6 +176,10 @@ class Agent(Module):
 
         if B_action_dependencies is not None:
             assert num_controls is not None, "Please specify num_controls if you're also using complex action dependencies"
+            if sampling_mode != "full":
+                raise ValueError(
+                    "Complex B action dependencies require sampling_mode='full'"
+                )
 
         if learn_A:
             assert pA is not None, "pA is required for A learning"
--- a/pymdp/agent.py
+++ b/pymdp/agent.py
@@ -167,11 +167,23 @@ class Agent(Module):
         learn_D: bool = False,
         learn_E: bool = False,
     ) -> None:
+        if action_selection not in {"deterministic", "stochastic"}:
+            raise ValueError(
+                "`action_selection` must be either 'deterministic' or 'stochastic'"
+            )
+        if sampling_mode not in {"full", "marginal"}:
+            raise ValueError("`sampling_mode` must be either 'full' or 'marginal'")
+
         if B_action_dependencies is not None:
             assert num_controls is not None, "Please specify num_controls if you're also using complex action dependencies"
 
         if learn_A:
             assert pA is not None, "pA is required for A learning"
+
+        if learn_C or learn_D or learn_E:
+            raise NotImplementedError(
+                "Modern Agent learning currently supports only learn_A and learn_B"
+            )
 
         # extract high level variables
         self.num_modalities = len(A)
@@ -789,7 +801,8 @@ class Agent(Module):
                 f"`empirical_prior` has {len(empirical_prior)} factor(s), expected {self.num_factors}"
             )
 
-        A = self.A
+        # Masking is per-call state, so never replace entries in self.A.
+        A = list(self.A)
         if mask is not None:
             for i, m in enumerate(mask):
                 o_vec[i] = m * o_vec[i] + (1 - m) * jnp.ones_like(o_vec[i]) / self.num_obs[i]
--- a/pymdp/agent.py
+++ b/pymdp/agent.py
@@ -175,11 +175,13 @@ class Agent(Module):
             raise ValueError("`sampling_mode` must be either 'full' or 'marginal'")
 
         if B_action_dependencies is not None:
-            assert num_controls is not None, "Please specify num_controls if you're also using complex action dependencies"
-            if sampling_mode != "full":
-                raise ValueError(
-                    "Complex B action dependencies require sampling_mode='full'"
-                )
+            is_canonical = B_action_dependencies == [[f] for f in range(len(B))]
+            if not is_canonical:
+                assert num_controls is not None, "Please specify num_controls if you're also using complex action dependencies"
+                if sampling_mode != "full":
+                    raise ValueError(
+                        "Complex B action dependencies require sampling_mode='full'"
+                    )
 
         if learn_A:
             assert pA is not None, "pA is required for A learning"
@@ -224,9 +226,11 @@ class Agent(Module):
 
         # flatten B action dims for multiple action dependencies
         self.action_maps = None
+        # canonical or empty dependencies don't require flattening
+        is_canonical = self.B_action_dependencies == [[f] for f in range(self.num_factors)]
         if (
-            policies is None and B_action_dependencies is not None
-        ):  # note, this only works when B_action_dependencies is not the trivial case of [[0], [1], ...., [num_factors-1]]
+            policies is None and self.B_action_dependencies is not None and not is_canonical
+        ):
             policies_multi_tup = control._construct_policies_tuple(
                 self.num_controls_multi,
                 self.num_controls_multi,
--- a/pymdp/inference.py
+++ b/pymdp/inference.py
@@ -21,6 +21,7 @@ from pymdp.algos import (
     hmm_smoother_from_filtered_colstoch,
 )
 from pymdp.maths import calc_vfe
+from pymdp import utils
 from jax import tree_util as jtu, lax
 from jax.experimental.sparse._base import JAXSparse
 from jax.experimental import sparse
@@ -180,6 +181,9 @@ def _run_one_step_inference(
 ) -> tuple[list[Array], list[Array] | Array]:
     curr_obs = _select_current_obs(obs, distr_obs)
     fpi_num_iter = 1 if method == EXACT_METHOD else num_iter
+    A_dependencies = utils.resolve_a_dependencies(
+        len(prior), len(A), A_dependencies
+    )
     qs = run_factorized_fpi(
         A,
         curr_obs,
--- a/pymdp/inference.py
+++ b/pymdp/inference.py
@@ -417,6 +417,17 @@ def update_posterior_states(
     if return_info and prior is None:
         raise ValueError("`prior` must be provided when `return_info=True`")
 
+    # Resolve default dependencies early so all methods and sequence inference receive complete mappings
+    num_factors = len(B) if B is not None else (len(prior) if prior is not None else 1)
+    num_modalities = len(A)
+    A_dependencies = utils.resolve_a_dependencies(
+        num_factors, num_modalities, A_dependencies
+    )
+    if B is not None:
+        B_dependencies = utils.resolve_b_dependencies(
+            num_factors, B_dependencies
+        )
+
     if method in SEQUENCE_METHODS:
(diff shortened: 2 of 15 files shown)


pyproject.toml
```diff
@@ -32,6 +32,7 @@ dependencies = [
     'seaborn>=0.11.1',
     'mctx>=0.0.5',
     'networkx>=3.3',
+    'numpyro>=0.19',
```
conorheins: We explicitly decided not to include numpyro in base dependencies so that it gets carried in indirectly if you use the `--modelfit` extra so it gets installed through `pybefit`

pymdp/learning.py
```diff
@@ -48,6 +53,8 @@ def update_obs_likelihood_dirichlet_m(
     relevant_factors = tree_map(lambda f_idx: qs[f_idx], dependencies_m)
 
     dfda = vmap(multidimensional_outer)([obs_m] + relevant_factors).sum(axis=0)
+    if support_mask is not None:
+        dfda = dfda * (support_mask > 0)
```
conorheins, line 68: [P1] This masks only the sufficient-statistic increment, not the returned expected matrix. dirichlet_expected_value clips zero concentrations to a positive epsilon, so E_qA becomes positive at forbidden cells; on the next online update those cells satisfy support_mask > 0 and start accumulating counts. I reproduced an identity-support A learning a forbidden [0, 1] entry on its second update; the B path has the same issue. Please keep a stable support mask through the returned model and add repeated-update assertions on both q and E outputs. Also note that support_mask > 0 currently has no BCOO rule, which is relevant to #446.

pymdp/utils.py
```diff
@@ -44,7 +44,11 @@ def norm_dist(dist: Array) -> Array:
     Array
         Normalized distribution.
     """
-    return dist / dist.sum(0)
+    denominator = dist.sum(0)
+    safe_denominator = jnp.where(denominator == 0, 1.0, denominator)
+    normalized = dist / safe_denominator
+    uniform = jnp.ones_like(dist) / dist.shape[0]
```
conorheins, line 57: [P2] This changes the return type for NumPy inputs: jnp.where produces a JAX array. Distribution.normalize stores that result, and its existing in-place __setitem__ then fails because JAX arrays are immutable. Please preserve the input array family here, or convert back to NumPy in Distribution.normalize, and cover normalize followed by mutation.

pymdp/agent.py
```diff
@@ -167,12 +167,28 @@ def __init__(
...
+        if action_selection not in {"deterministic", "stochastic"}:
+            raise ValueError(
+                "`action_selection` must be either 'deterministic' or 'stochastic'"
+            )
+        if sampling_mode not in {"full", "marginal"}:
+            raise ValueError("`sampling_mode` must be either 'full' or 'marginal'")
+
         if B_action_dependencies is not None:
             assert num_controls is not None, "Please specify num_controls if you're also using complex action dependencies"
+            if sampling_mode != "full":
```
conorheins: [P2] Non-None does not mean complex. The canonical mapping [[0]] is semantically identical to omitting B_action_dependencies and works with marginal sampling, but this guard rejects it. The added test currently reinforces that regression because it uses a self-only mapping rather than a coupled action mapping. Please distinguish genuinely coupled mappings or bypass flattening for the canonical case.

pymdp/inference.py
```diff
@@ -180,6 +181,9 @@ def _run_one_step_inference(
 ) -> tuple[list[Array], list[Array] | Array]:
     curr_obs = _select_current_obs(obs, distr_obs)
     fpi_num_iter = 1 if method == EXACT_METHOD else num_iter
+    A_dependencies = utils.resolve_a_dependencies(
```
conorheins, line 184: [P2] Resolving dependencies only in the one-step helper leaves the public contract inconsistent: FPI now accepts A_dependencies=None, but MMP and VMP still pass None into sequence inference and fail; omitted B_dependencies fails there as well. Please resolve both mappings before method dispatch, or inside the sequence path, and extend the regression to MMP and VMP.

test/test_likelihoods.py
```diff
@@ -0,0 +1,11 @@
+"""Direct import coverage for the shipped NumPyro likelihood module."""
+
+import numpyro
```
conorheins: [P2] Once the direct NumPyro dependency is removed per the packaging policy above, this unconditional import makes a minimal base-package test collection fail. CI currently masks that because the test dependency group installs pybefit, which supplies NumPyro transitively. Please make this an optional or modelfit smoke test, for example with importorskip and the existing reason, rename the declared-dependency wording, or fold it into the pybefit integration test.

conorheins (changes requested):
Thanks for the targeted hardening work. I am requesting changes on the current head for one correctness blocker and several API and packaging-contract regressions:

- Structural-zero masks are not stable across repeated online updates; the returned expected matrices re-enable forbidden entries.
- norm_dist changes NumPy-backed Distribution data into immutable JAX arrays.
- Canonical explicit action dependencies are rejected as if they were complex.
- Default dependency resolution still fails in MMP and VMP.
- NumPyro should remain optional through the modelfit and pybefit path, and the likelihood smoke test should reflect that contract.

The branch also needs rebasing onto main at 5848360. I reproduced the rebase: the only conflict is the dependency tail in pyproject.toml. Please keep opt_einsum and jaxtyping from #426 and omit NumPyro; the other commits replay cleanly. Under that minimal dependency policy, the changed-test subset gives 59 passes plus one collection error in test_likelihoods because NumPyro is imported unconditionally; the normal CI group masks that by installing it through pybefit.

Please rerun CI after rebasing and addressing the inline findings.

docxology:
Thanks @conorheins for the thorough review. I've rebased the branch onto `upstream/main` (`5848360`) and addressed all inline findings:

1. **Structural-zero stability across online updates (`pymdp/learning.py`)**:
   - Both `update_obs_likelihood_dirichlet_m` and `update_state_transition_dirichlet_f` now explicitly mask both the concentration parameters (`qA` / `qB`) and returned expected matrices (`E_qA` / `E_qB`) with `_mask_support`, renormalizing strictly over valid support columns.
   - Added repeated online update test coverage verifying that passing `E_qA` / `E_qB` back into consecutive updates preserves zero support without re-enabling forbidden entries across iterations.

2. **Array family preservation in `norm_dist` & `Distribution.normalize` (`pymdp/utils.py`, `pymdp/distribution.py`)**:
   - `norm_dist` now preserves NumPy array inputs without converting them to immutable JAX arrays via `jnp.where`.
   - `Distribution.normalize()` explicitly ensures data remains a mutable `np.ndarray`.
   - Added test coverage in `test/test_distribution.py` and `test/test_utils_jax.py` verifying in-place item mutation post-normalization.

3. **Canonical vs complex action dependencies in `Agent` (`pymdp/agent.py`)**:
   - `Agent.__init__` now distinguishes canonical/per-factor action dependencies (`[[f] for f in range(num_factors)]`) from coupled mappings. Canonical dependencies are accepted under `sampling_mode="marginal"` and bypass multi-action policy flattening.
   - Added test coverage accepting canonical dependencies with marginal sampling and rejecting genuinely coupled dependencies.

4. **Default dependency resolution across inference methods (`pymdp/inference.py`)**:
   - `update_posterior_states` now resolves default `A_dependencies` and `B_dependencies` at entry prior to dispatching to one-step or sequence inference.
   - Added multi-factor test coverage verifying omitted dependencies work consistently across FPI, MMP, and VMP.

5. **Optional NumPyro / pybefit test collection**:
   - Replaced unconditional imports in `test/test_likelihoods.py` and `test/test_tmaze_recoverability.py` with explicit `pytest.importorskip` for `numpyro` and `pybefit`.
   - Clean base test suite runs without requiring the `--modelfit` extra.

Full test suite verified locally: 318 passed, 2 skipped, 78 subtests passed.

conorheins:
@docxology one more thing for the rebase: #452 just landed and fixes the flake you worked around in `test_inference_jax.py` properly -- both sides of that comparison run in float64 now, so the error sits about six orders of magnitude inside the default tolerance. That means the `atol=1e-4, rtol=1e-4` on `test_fixed_point_iteration_multistate_multiobs` can go; 1e-4 is loose enough to hide a real regression in the FPI path. Take main's version of that hunk when you rebase.

conorheins:
Thanks @docxology -- went through the head (`6af9f22`). The masking does what it says for the A path and for updated B factors: on main a forbidden `A` cell climbs 0.286 -> 0.444 -> 0.545 over three consecutive updates, and on your branch it stays exactly 0. Two things left, plus one that changed underneath you.

**1. The skip branch still leaks.** `factors_to_update != "all"` sends skipped factors through the `lax.cond` false branch, `lambda: (pB_f, dirichlet_expected_value(pB_f))`, which never sees `_mask_support`. `dirichlet_expected_value` clips to `MINVAL`, so forbidden transitions come back at 2.2e-16 -- and since the support mask is derived from the `B` you pass in, the next update that *does* touch that factor treats them as supported:

```
update with factors_to_update=[0]   ->  E_qB[1] forbidden cell = 2.2e-16
feed that back in, update [1]       ->  E_qB[1] forbidden cell = 0.247
```

Same on main, so it's the tail of the original finding rather than something you introduced. Either mask the false branch too, or have it return the `B_f` it was handed instead of recomputing the expected value. Needs a repeated-update test with `factors_to_update=[0]` -- the current ones only cover `"all"`.

**2. The canonical check is narrower than "not coupled".** `is_canonical` is literal equality with `[[f] for f in range(len(B))]`, so `[[1], [0]]` -- one action per factor, nothing coupled, just permuted -- gets `ValueError: Complex B action dependencies require sampling_mode='full'`. That case works on main today. The condition you want is "each factor depends on at most one action and no action is shared", not equality with the identity list.

**3. #451 landed on the same lines.** Main now warns instead of raising when flattened action deps meet `sampling_mode="marginal"`, and it removed the `self.sampling_mode = "full"` assignment in the flatten branch -- that assignment was dead, the constructor arg overwrote it right after. Your branch reintroduces both. Take main's version on the rebase; if you think raising is right, let's settle that on #434 rather than carry both.

You're 6 commits behind now (merge base is #426), so that plus the tolerance revert above is the rebase.

PR #449, opened 2026-08-25
feat(inference): completed the optimized infer_states variants (FPI, VMP, MMP) from the paper https://arxiv.org/abs/2607.17855

NikolaPizurica:
## Summary

This PR extends the optimized state-inference work from #315 (FPI hybrid / block-diagonal / end-to-end padded methods) in two directions:

1. **VMP and MMP variants** of all six optimized method families (hybrid, clustered hybrid, hybrid block, clustered hybrid block, end-to-end padded, clustered end-to-end), for sequence-based inference with **trivial `B_dependencies`** (each factor depends only on its own state).
2. **Clustered variants for FPI** (clustered hybrid, clustered hybrid block, clustered end-to-end), completing the same six-method roster for FPI.

All optimized routines are validated against the reference implementations (`update_posterior_states` with `method='fpi' | 'vmp' | 'mmp'`) via `jnp.allclose` across generated agent specs, batch sizes, and sparse/dense `A`.

## What's added

**`pymdp/algos.py`** — entry points mirroring the #315 conventions (precomputed log-likelihoods in, posteriors out):
- `run_vmp_hybrid`, `run_vmp_end2end_padded`, `run_vmp_clustered_end2end`
- `run_mmp_hybrid`, `run_mmp_end2end_padded`, `run_mmp_clustered_end2end`
- `run_factorized_fpi_clustered_end2end` (the "partially padded clustered" method from the accompanying research code)
- shared helpers: `prep_lnB_and_ln_prior` / `prep_B_and_ln_prior`, `pad_qs_and_compute_vmp_msgs` / `pad_qs_and_compute_mmp_msgs`, `mirror_gradient_descent_step_combined` (a re-parametrization of `mirror_gradient_descent_step` taking pre-summed messages), and the clustered marginal-likelihood machinery (`get_qs_clusters`, `compute_qL_marginals_clustered`, `compute_qL_all_clustered`, `all_marginal_log_likelihood_end2end`, `all_marginal_log_likelihood_clustered`)

**`pymdp/maths.py`**
- padded message primitives: `get_vmp_fw_msgs_padded` / `get_vmp_bw_msgs_padded` and `get_mmp_fw_msgs_padded` / `get_mmp_bw_msgs_padded` (docstrings note these are the trivial-`B_dependencies`, all-valid specializations of `get_vmp_messages` / `get_mmp_messages`), plus the shared `deconstruct_msgs_padded`
- per-cluster log-likelihood family: `compute_log_likelihoods_per_cluster`, `deconstruct_log_likelihoods_per_cluster`, `compute_log_likelihoods_end2end_per_cluster`, `compute_log_likelihoods_block_diag_clustered`

**`pymdp/utils.py`**
- sequence helpers: `init_B_from_spec`, `get_sample_action_seq`, `get_sample_state`, `get_obs_seq_from_actions`, `pad_individual_Bs`, `preprocess_B` (docstring cites its parent, `inference._condition_transitions_on_actions`, all-valid-actions path)
- modality clustering: `get_A_dep_clusters`, `apply_padding_per_cluster`, `apply_A_end2end_padding_per_cluster`, `apply_obs_end2end_padding_per_cluster`, `agglomerative_min_structural_zeros`, `kneedle_max_distance_to_chord`, `prep_clustered_block_data`

**Notebooks & tests**
- `examples/advanced/infer_states_optimization/`: `methods_test_fpi.ipynb` (extended to all six FPI methods), new `methods_test_vmp.ipynb` and `methods_test_mmp.ipynb`; all three registered in `test/notebooks/ci_notebooks.txt` and linked from the docs tutorials index
- `test/`: `test_infer_states_fpi_optimized.py` (extended to six methods), new `test_infer_states_vmp_optimized.py` and `test_infer_states_mmp_optimized.py`, each comparing all methods (JIT-compiled, dense and sparse) against the reference implementation

## Changes to existing code (backward compatible)

Three #315 helpers were generalized in place so the FPI and VMP/MMP paths share one implementation; in each case the previous behavior is the default or provably identical for the previous input shapes, and the existing FPI test suite passes unchanged:
- `deconstruct_lls`: new `has_time_axis: bool = False` keyword
- `apply_obs_end2end_padding_batched`: pads the trailing axis for arbitrary leading axes (`(batch,)` before, now also `(batch, T)`)
- `concatenate_observations_block_diag`: concatenates on `axis=-1` (identical to `axis=1` for the 2-D FPI inputs)

## Notable API changes (reviewer attention welcome)

- **Rename (typo fix):** `compute_log_likelihood_per_modality_end2end2_padded` → `compute_log_likelihood_per_modality_end2end_padded`, with call sites updated (test + notebook). Happy to add a deprecation alias for the old name if preferred.
- **Renames within this PR's own additions** during MMP integration (`get_fw_msgs_padded` → `get_vmp_fw_msgs_padded`, etc.) — these never shipped in a release.
- **File renames** for symmetry with the new siblings: `test_infer_states_optimized.py` → `test_infer_states_fpi_optimized.py`, `methods_test.ipynb` → `methods_test_fpi.ipynb` (CI notebook manifest, Colab badge, and docs links updated accordingly).

## Notes for reviewers

- The VMP/MMP routines assume trivial `B_dependencies` and all-valid observations/transitions; this is documented in the docstrings. `run_vmp` / `run_mmp` remain the general implementations and are untouched.
- `compute_qL_all_clustered` intentionally initializes with `jnp.zeros(1)` (broadcastable over ragged per-factor shapes) and must not be folded into `compute_qL_all` — documented in its docstring.
- The MMP backward-message normalization keeps the research implementation's `+ 1e-16` guard (vs. `clip(min=MINVAL)` in `get_mmp_messages`); equivalent for well-formed transition tensors, preserved to match the validated research numerics.

## Validation

- `pytest test/` green locally (FPI, VMP, and MMP suites; dense + sparse, batch sizes 1 and 4).
- All three notebooks execute end-to-end with every method matching the reference posteriors via `allclose`.


conorheins (changes requested):
Thanks @NikolaPizurica, and congrats on getting the paper out -- really glad to have VMP/MMP versions of the optimized methods. Short version: the numerics check out, there's one real correctness bug around invalid actions, and the PR is a lot bigger than it needs to be because VMP and MMP are duplicated at every layer. Agent integration I'll handle separately, so that's out of scope here.

## Numerical correctness

I ran all six families (VMP and MMP) plus the new clustered FPI against `update_posterior_states` on cases the current tests don't cover: T = 0, 1, 3 and 8; `num_iter` 1 and 16; batch 1; factors with 1, 3 and 4 controls; non-uniform priors; deterministic (permutation and shift) `B`; `A` with 60-90% structural zeros including sparse BCOO; observations that are impossible under `A`; ragged state/obs dims; and several clusters.

- Everything matches to float32 precision (mostly ~1e-7, a few e-6 with deterministic `B`), including the new clustered FPI.
- VMP deviates by ~3e-5 when `B` has zero-probability transitions or an observation has zero likelihood. In float64 it agrees to 1e-13, so this is float32 amplification rather than a logic difference -- but it's above your tests' `ATOL`, so worth knowing if you add deterministic-`B` cases.

## The `-1` action bug

`preprocess_B` re-implements `inference._condition_transitions_on_actions` but drops its invalid-action handling, so `b[..., -1]` silently selects the *last* action. The reference path masks those transitions out instead (`transition_valid_mask`, built from `past_actions >= 0`). With a `-1` in `past_actions`, every family is off by up to 0.3 (VMP) and 0.1 (MMP), with no error.

The reference relies on this: for windowed sequence inference, `rollout()` starts every episode with `action_0 = -jnp.ones(...)` and uses `valid_steps` and action-validity masks to ignore the padded slots. So these methods can't stand in for the reference until they handle it. Please reuse `_condition_transitions_on_actions` (on the unpadded `B`, then pad) and accept the same obs/transition valid masks the reference takes -- or at minimum raise on negative actions for now.

**Tests for this.** The current tests can't hit it, since `get_sample_action_seq` only draws valid actions. Please add tests comparing each family against `update_posterior_states` with `-1` in `past_actions` -- one at t=0 the way `rollout()` produces it, and one mid-sequence -- plus padded windows via `valid_steps` if you add mask support.

## Eligibility checks as plain functions

Non-trivial `B_dependencies` currently fails with `ValueError: too many values to unpack (expected 4)`. Could you add small standalone functions that check the assumptions each layout makes (trivial `B_dependencies`, plus valid actions if you don't add mask support) and raise a clear error, and call them from the entry points? Having them as plain functions also makes these methods easy to plug into the Agent later.

## Duplication

Roughly three quarters of the diff is duplicated code or stored notebook output:

- **VMP vs MMP.** `run_vmp_hybrid` and `run_mmp_hybrid` differ in exactly two lines -- which prep function and which message function they call -- and the same is true of the end2end and clustered pairs, `pad_qs_and_compute_{vmp,mmp}_msgs`, and `prep_{lnB,B}_and_ln_prior` (~530 duplicated lines). The reference implementation already has a pattern for this: `update_marginals(get_messages, ...)` is shared, and `run_vmp`/`run_mmp` just pass in `get_vmp_messages`/`get_mmp_messages`. Could the optimized runners follow the same pattern -- one hybrid, one end2end and one clustered runner, each taking the padded message functions (and the log-or-not prep) as arguments? That also leaves one entry point per layout with a common signature.
- **Near-copies of existing functions.** `mirror_gradient_descent_step_combined` and `compute_qL_all_clustered` can both go. I tried it locally: calling the existing `mirror_gradient_descent_step(tau, ln_A, lnB_past_and_future, 0.0, ln_q)` at the six call sites, and changing `compute_qL_all`'s initializer to `jnp.zeros(1)` (which covers the ragged clustered case) and using it at the two clustered call sites. Your three optimized test suites still pass (18 tests, 21 subtests) and posteriors are unchanged across all the cases above.
- **Tests.** The VMP and MMP test files are 495 lines each and differ only in the class name -- one file parametrized over the method, please.
- **Notebooks.** The VMP and MMP notebooks have identical code apart from the method name, and ~85% of each file is stored output from cells that end on a bare expression (each echoes a ~32 KB tuple). One notebook with a method switch, with those cells ending in `;`, would drop about 9k lines.

## Smaller things

- `init_B_from_spec`, `get_sample_action_seq` and `get_sample_state` use NumPy's global RNG, unseeded. We don't use NumPy RNG in the JAX codebase, so please take JAX keys instead -- that also makes the tests deterministic. They're test fixtures, so I'd move them out of `pymdp/utils.py` too.
- `init_A_and_D_from_spec` always returns a uniform `D`, and the sequence tests all use T=8, `num_iter=8`, two controls per factor and random dense `B`. Please add cases for non-uniform priors, deterministic `B`, short sequences and other iteration counts. Happy to share my harness if it's a useful starting point.
- `compute_log_likelihood_per_modality_end2end2_padded` shipped in 1.0.4, so please keep the old name as a deprecated alias rather than renaming outright.

Suggested order: (1) the `-1` fix and its tests, plus the eligibility checks; (2) the VMP/MMP consolidation, reusing the existing functions, and the parametrized tests; (3) the single notebook. Separate PRs are fine if that's easier. Thanks!

NikolaPizurica:
Hi @conorheins,
Thank you for the detailed review. It sounds like this shouldn't be too difficult, but at the moment I have a ton of other things going on so I might not be able to do it right away. I will let you know as soon as I find some space to focus on this properly.

